PR #7, opened 2022-10-05
Added Obliteration png and ico files.

Mou-Ikkai:
Added the new Obliteration program icon! Path is src/resources/

obliteration-icon


ultimaweapon:
Cool!

PR #11, opened 2022-10-07
Added Github Releases and PR actions.

Mou-Ikkai:
Please for the love of god, when merging this, use the Squash and Merge option for commit history XD
Either way, when a push/commit is added to the Main branch, a Github Release of a Linux and Windows executable is made and published! When a different branch or a PR is used, it instead makes the Linux and Windows executables, and uploads Artifacts you can download from the build page!

Publish/Release example:
image

External Build example: ([You can check the builds it did just for this PR! :3](https://github.com/ultimaweapon/obliteration/actions/runs/3205042849))
image


ultimaweapon (changes requested):
Thanks for the PR. There are some issues with `.github/workflows/external.yml` and `.github/workflows/main-publish.yml`:

- We can't use `windows-latest` because it is using Visual Studio 2022. The reason is Qt 6.2 was built with 2019. If we built Obliteration with 2022 it will cause incompatible Visual C++ runtime library because our using 2022 while Qt using 2019.
- I find it hard to read and maintain when we merging `steps` for both Windows and Linux. We should merge only if both can share most of steps without conditional like `matrix.os == 'windows-latest'`.
- If we create a release on every push to `main` branch our release page will be spammed because our `main` branch is a development branch. We should create a release only when we push a new tag instead, which we should not implement this for now because it is likely the build process will be different when we doing a first release.
- We should not use `ubuntu-latest` because Vulkan SDK that we install in the build step is for 20.04. If we use `ubuntu-latest` and GitHub decided to upgrade it to newer version our build step will be broken.
- The binaries that we export from Ubuntu will only work for Ubuntu. It will not work on other Linux. There is a solution for this like Flatpak, Snap or AppImage but we should not implement this for now because it is likely our build process will be different when we release the first version.

So I think we should stick with the current `.github/workflows/ci.yml` for now.

Mou-Ikkai:
@ultimaweapon I'll fix the issues you said above, also, to avoid build spams, I'm thinking I'll use a Schedule run to make it run every monday at 00:00 UTC. So... Weekly releases. (For which, seems good with the rate this project is worked on)

ultimaweapon:
We should dedicated a release page for only version that we intentional release to the public like when the first game is bootable. With this way people who have stared on this project will be notified about the progress that we want to announce to the public. If we create a release for a version that is not ready for the public yet it will going to spam people who have stared with unmeaningful notification.

We already have a latest develop build on every push so people who are interested in the latest development version can grab from there.

Mou-Ikkai:
Alright, I'll have artifacts be made for all pushes, and when a new tag is made, the release is created. Sound good?

Mou-Ikkai:
Tags match the format: v*.*.*
For building. This allows us to make tags without accidentally making builds.

.github/workflows/main-publish.yml
```diff
@@ -0,0 +1,99 @@
...
+        run: cmake --build build --config Release
+        
+      - name: Export artifacts
+        run: cmake --install build --prefix dist
+        
+      - name: ZIP Obliteration
+        uses: TheDoctor0/zip-release@main
+        with:
+          type: 'zip'
+          path: './build/src/*'
```
ultimaweapon, line 88: We should not use outputs from this directory because it contains a lot of unnecessary files for the user. We should use `dist` instead.

.github/workflows/build.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: cmake --build build --config Release
+        
+      - name: Export artifacts
+        run: cmake --install build --prefix dist
+        
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: Obliteration-${{ env.version }}-${{ matrix.release_suffix }}
+          path: build/src/
```
ultimaweapon, line 91: Same here.

ultimaweapon (changes requested):
I forgot that we use dependencies from external repositories for Ubuntu. That mean the output binaries will not work on user machine. So better to not upload the binary for Ubuntu unless we use Flatpak or similar technology.

Mou-Ikkai:
@ultimaweapon Done! Let's see if it works.

ultimaweapon:
Seem like you forgot to remove Ubuntu binary from the release & artifact.

Mou-Ikkai:
> Seem like you forgot to remove Ubuntu binary from the release & artifact.

Well, I will keep the build, but not the upload. How about that? Just so we can make sure it doesn't break the building on Ubuntu

Mou-Ikkai:
@ultimaweapon How about now?

Mou-Ikkai:
image
EXE works.

ultimaweapon:
Thanks for testing PKG installation. I always wonder if it going to work with other PKG or not.

ultimaweapon (approved):

PR #19, opened 2022-10-27
Update to show 64 bit on builds.

Mou-Ikkai:
We compile 64-bit builds because the PS4 uses a 64 Bit CPU... Yet the file name was ending with win32, I changed it to win64


ultimaweapon (approved):
Actually the word `win32` mean legacy but still widely used [Windows API](https://en.wikipedia.org/wiki/Windows_API). But I guess rename it to `win64` should be better for non-technical people.

PR #34, opened 2022-11-01
Expect a Zlib header when decompressing.

Mou-Ikkai:
Decompress has a setting specifically for ZLib files, testing to see if this works.

### DIFF
--- a/src/pfs/src/pfsc.rs
+++ b/src/pfs/src/pfsc.rs
@@ -1,4 +1,4 @@
-use flate2::read::ZlibDecoder;
+use flate2::FlushDecompress;
 use std::cmp::min;
 use std::error::Error;
 use std::fmt::{Display, Formatter};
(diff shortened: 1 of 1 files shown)


src/pfs/src/pfsc.rs
```diff
@@ -100,13 +100,13 @@ impl<F: Read + Seek> Reader<F> {
...
-            let mut compressed = new_buffer(size as usize - 2);
+            let mut compressed = new_buffer(size as usize);
 
-            self.file.seek(SeekFrom::Start(offset + 2))?;
+            self.file.seek(SeekFrom::Start(offset))?;
             self.file.read_exact(&mut compressed)?;
 
             // Decompress.
-            let mut deflate = flate2::Decompress::new(false);
+            let mut deflate = flate2::Decompress::new(true);
```
ultimaweapon, line 109: You need to change this to https://docs.rs/flate2/1.0.24/flate2/read/struct.ZlibDecoder.html. See [here](https://github.com/ultimaweapon/obliteration/blob/main/src/pup/src/entry.rs#L233) for how we this with PUP.

ultimaweapon (changes requested):

PR #39, opened 2022-11-01
ZLib implementation test.

Mou-Ikkai:
Haha, I don't know C++, I'm gonna try to learn.

### DIFF
--- a/src/pfs/src/pfsc.rs
+++ b/src/pfs/src/pfsc.rs
@@ -1,4 +1,4 @@
-use flate2::FlushDecompress;
+use flate2::read::ZlibDecoder;
 use std::cmp::min;
 use std::error::Error;
 use std::fmt::{Display, Formatter};
--- a/src/pfs/src/pfsc.rs
+++ b/src/pfs/src/pfsc.rs
@@ -100,14 +100,14 @@ impl<F: Read + Seek> Reader<F> {
             buf.fill(0);
         } else {
             // Read compressed.
-            let mut compressed = new_buffer(size as usize - 2);
+            let mut compressed = new_buffer(size as usize);
 
-            self.file.seek(SeekFrom::Start(offset + 2))?;
+            self.file.seek(SeekFrom::Start(offset))?;
             self.file.read_exact(&mut compressed)?;
 
             // Decompress.
-            let mut deflate = flate2::Decompress::new(false);
-            let status = match deflate.decompress(&compressed, buf, FlushDecompress::Finish) {
+            let mut deflate = ZlibDecoder::new(buf);
+            let status = match deflate.read_to_end(&mut compressed) {
                 Ok(v) => v,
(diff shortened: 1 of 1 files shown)


Mou-Ikkai:
@ultimaweapon I broke it-

src/pfs/src/pfsc.rs
```diff
@@ -100,14 +100,14 @@ impl<F: Read + Seek> Reader<F> {
...
+            let mut compressed = new_buffer(size as usize);
 
-            self.file.seek(SeekFrom::Start(offset + 2))?;
+            self.file.seek(SeekFrom::Start(offset))?;
             self.file.read_exact(&mut compressed)?;
 
             // Decompress.
-            let mut deflate = flate2::Decompress::new(false);
-            let status = match deflate.decompress(&compressed, buf, FlushDecompress::Finish) {
+            let mut deflate = ZlibDecoder::new(buf);
```
ultimaweapon, line 109: `ZlibDecoder::new` accept [Read](https://doc.rust-lang.org/std/io/trait.Read.html) that provides Zlib data. That mean you should pass `compressed` as a slice of `u8` to it because a slice of `u8` implements [Read](https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-for-%26%5Bu8%5D).

`compressed` has a type of `Vec<u8>`. So you need to figure out how to get a slice of `u8` from a vec of `u8`. Good luck :)

ultimaweapon (changes requested):
Actually the code you are currently working on is Rust, not C++. I'll guide you one by one until you have fixed all required code.

Mou-Ikkai:
Made a sliced var as I'm unsure if regular compressed is still needed.

Mou-Ikkai:
...Seems to be unhappy with what I did .w.

Mou-Ikkai:
@ultimaweapon This hurts xD

Mou-Ikkai:
So, explicitly, WHY do we have to use Zlib? The Decompress option had a toggle for ZLib headers, after all.

ultimaweapon:
You right. Not sure if it have the same functionality. Can you take a look on Zlib source to see if it actually do nothing?

Mou-Ikkai:
> You right. Not sure if it have the same functionality. Can you take a look on Zlib source to see if it actually do nothing?

Well, I pushed the change for Deflate to have the Zlib header flag enabled, and it decoded the PKG and extracted the game with no sadness/errors.

Mou-Ikkai:
> You need to start the game to see if it works because the affected file is `pfs_image.dat`, which will be using only when starting the game.

...I did. Amplitude. Besides from Kyty's same error of not supporting a function yet (this happens just on base Kyty, not implemented), it works.

ultimaweapon:
If you have reached unknown instruction error then we can assume it works.

Mou-Ikkai:
> If you have reached unknown instruction error then we can assume it works.

Correct!

Issue #51, opened 2022-12-21
Add MacOS support for Intel


bayedieng:
This will potentially be very difficult to do with decent performance but it would be quite nice to see macos and arm support (as that's what apple is increasingly moving towards). In regards to CPU emulation, the Unicorn Engine could be used but am personally unfamiliar with its performance and Licensing could be an issue since it's GPL. There is also the option of Rosetta but it is unclear how long apple will support it.

Regarding GPU emulation, since the core emulation is written in Rust rendering with WGPU might be a good option.

ultimaweapon:
Thanks for the information. Actually it is possible for ARM and macOS because we reassemble the whole executable and its dynamic modules. That mean we can replace the assemble part to output ARM instructions instead of x86. The problem is we need someone to work on this part because we don't have enough capacity.

bayedieng:
> The problem is we need someone to work on this part because we don't have enough capacity.
That is something that I could work on at some point given time or if no one else begins work on it first. 

In regards to gpu emulation, would wgpu be viable for this project?

ultimaweapon:
That will be awesome! I'll let you know when I finished the abstraction for the assemble part so you can create an implementation for ARM.

For WGPU I did not take a look yet. We have not reached the GPU part yet so I really have no idea how the PS4 GPU works in detail. The only thing I know is it uses unified memory, which is the reason why I chose to emulate system calls and reassemble the whole binary.

ultimaweapon:
In the mean time do you mind working on macOS CI and making the code buildable on macOS? Currently GitHub Actions does not provide M1 machine but we can use Intel until M1 is available.

bayedieng:
Alright, will attempt to get initial macos build support. I have an m1 mac though, not Intel.  I will not be able to test the Intel build locally but it should still work the same.

ultimaweapon:
@bayedieng cloud you check size of memory page on your macOS with M1 (e.g. `getconf PAGESIZE`)? I already got it from https://cpufun.substack.com/p/more-m1-fun-hardware-information but just want to make sure.

ultimaweapon:
Thanks!

bayedieng:
> Are you interested?
Yeah sure.

ultimaweapon:
@Mou-Ikkai do you want to join the organization? I already know the answer but just want to make sure :)

Mou-Ikkai:
@ultimaweapon That's gonna be a YESSSSSSSSS

[...]

PR #52, opened 2022-12-23
Add CI for macos

bayedieng:
Adding CI support is the first towards supporting macos. See #51.


Mou-Ikkai:
Keep checking Checks. Seems you have some Yaml errors.

bayedieng:
It builds successfully on my mac. It seems to be failing due to a filesystem error as a result of c++17 not being supported by the clang compiler.

ultimaweapon (approved):
Looking good. Thanks!

PR #77, opened 2023-01-21
Initializes memory manager

ultimaweapon:
The idea of memory manager is it will manage all of page allocation by the PS4 app. The reason we need this because the size of memory page on PS4 is not the same as PC (16K vs 4K). Each page also have additional protection for GPU while the PC does not have this.


bayedieng:
Looks good! Although the RAM usage requirement  does seem quite high. Is there anyway to mitigate this?

bayedieng (approved):

PR #79, opened 2023-01-21
Add MacOS build files.

### LINKED ISSUE #74: Produce app bundled for macOS on CI
So macOS users do not need to build from sources.

Mou-Ikkai:
Adds release files for MacOS, also fixed CMake caching path as its different than Linux.

Should fix #74

### DIFF
--- a/.github/workflows/ci-mac.yml
+++ b/.github/workflows/ci-mac.yml
@@ -19,10 +19,17 @@ jobs:
       id: cache-cmake
       with:
         path: |
-          /home/runner/work/obliteration/obliteration/build
-          /home/runner/work/obliteration/obliteration/src/target
+          /Users/runner/work/obliteration/obliteration/build
+          /Users/runner/work/obliteration/obliteration/src/target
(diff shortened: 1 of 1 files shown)


ultimaweapon (changes requested):
Nah it is more complicated than this. Take a look here for what is an App bundled on macOS: https://en.wikipedia.org/wiki/Bundle_(macOS)

bayedieng:
I myself, am not quite familiar with MacOS app bundling but do know some sort of certification from Apple for an official you could have an uncertified build and allow it in security settings as an end-user like @Mou-Ikkai did but I'm guessing we want certified builds. Certifying the binary requires signing into Apple so I'm not sure how that's going to work.

ultimaweapon:
We cannot have a signed version for now because it need some money... so let's do an unsigned version for now until we are at the stage to open for a donation.

ultimaweapon:
This should be a good starting point: https://stackoverflow.com/questions/53560288/how-to-create-a-macos-app-bundle-with-cmake

PR #86, opened 2023-01-24
Revises memory manager

ultimaweapon:
With this version we don't need a custom allocator anymore because we can map `mmap` and `munmap` to the host API directly.


bayedieng:
It's probably best to have system calls in Linux and macOS targeted differently since macOS tends to have some slight variations to POSIX.

Mou-Ikkai:
Also, I would say to keep the ReadME memory requirement. The PS4 has 8GB of RAM, which we can't just make that magically smaller.

src/kernel/src/memory/mod.rs
```diff
@@ -1,61 +1,53 @@
...
+        // granularity because it is always multiply of page size, which is a correct value.
+        let (page_size, allocation_granularity) = Self::get_memory_model();
 
         if page_size > Self::VIRTUAL_PAGE_SIZE || (Self::VIRTUAL_PAGE_SIZE % page_size) != 0 {
+            // If page size is larger than PS4 we will have a problem with memory protection.
+            // Let's say page size on the host is 32K and we have 2 adjacent virtual pages, which is
+            // 16K per virtual page. The first virtual page want to use read/write while the second
+            // virtual page want to use read-only. This scenario will not be possible because those
+            // two virtual pages are on the same page.
             panic!("Your system are using unsupported page size.");
```
Mou-Ikkai, line 31:
```suggestion
            panic!("Your system is using an unsupported page size.");
```

Mou-Ikkai (commented):
Small grammar change.

ultimaweapon:
@bayedieng what do you mean?

ultimaweapon:
@Mou-Ikkai the memory usage depends on the game. It is very unlikely to reach 8 GB. Some simple 2D games may use just a couple hundred megabytes.

bayedieng:
> @bayedieng what do you mean?

You are targeting ```unix``` for your syscall functions which linux and macOS are both apart of. The issue here is that macOS sometimes doesn't support the same flag/variable for the same syscall so it's better to either target linux and macos individually instead of unix.

ultimaweapon:
We can use the same for most POSIX APIs, including this case. It is quite rare to have a separated API between Linux and macOS.

bayedieng:
Yeah, I'm fine with using unix for most syscalls, in this case however, I don't think the sysconf will work because macOS does does not support the flag you are using.

ultimaweapon:
Interesting. I tried to find some information about this but did not find how to get page size on macOS.

ultimaweapon:
I believe it is working according to this issue: https://github.com/rust-lang/miri/issues/216

ultimaweapon:
It works on the iPhone, which means it should work on macOS too: https://developer.apple.com/forums/thread/47532

bayedieng:
I see Apple's online syscall  manpages seem to be outdated. Should be good to merge then.

bayedieng (approved):

PR #87, opened 2023-01-25
implement our own kernel debug parser

### LINKED ISSUE #16: Parsing kernel debug arguments manually
Right now we use `serde_yaml` which is quite overkill. So it is better to parse it manually because the file structure that we are using is very simple.

bayedieng:
This commit implements a rather simple parser for the .kernel-debug file and should close #16


ultimaweapon (changes requested):
I think the parser should be in the kernel, not in the util; because it is kernel specific and does not use by other crates.

ultimaweapon (approved):

PR #89, opened 2023-01-26
fix for windows path

### LINKED ISSUE #16: Parsing kernel debug arguments manually

bayedieng:
This should close #16. Someone would have to test it on Windows first.

### DIFF
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -174,8 +174,16 @@ fn mount_pfs<G: AsRef<Path>>(fs: &Fs, game: G) -> bool {
     // Open PFS image.
     let mut path = game.as_ref().to_path_buf();
 
-    path.push("uroot");
-    path.push("pfs_image.dat");
+    #[cfg(unix)]
+    {
+        path.push("uroot");
+        path.push("pfs_image.dat");
+    }
+    #[cfg(windows)]
+    {
+        path.push("\\uroot");
+        path.push("\\pfs_image.dat");
+    }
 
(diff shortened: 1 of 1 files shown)


src/kernel/src/main.rs
```diff
@@ -174,8 +174,16 @@ fn mount_pfs<G: AsRef<Path>>(fs: &Fs, game: G) -> bool {
...
-    path.push("pfs_image.dat");
+    #[cfg(unix)]
+    {
+        path.push("uroot");
+        path.push("pfs_image.dat");
+    }
+    #[cfg(windows)]
+    {
+        path.push("\\uroot");
+        path.push("\\pfs_image.dat");
```
ultimaweapon, line 185: This does not look right. `PathBuf` should already handled this for us.

ultimaweapon (changes requested):

src/kernel/src/main.rs
bayedieng: When @Mou-Ikkai used the parser it pushed uroot and pf_image.dat as a unix path for some reason, even though the Windows path parses correctly.

ultimaweapon: The problem is it was working correctly before we switched the argument parser. My idea is we should write a test for the argument parser to see if it does what we expected.

bayedieng:
Legitimately have no idea

Issue #73, opened 2023-01-20
Add AArch64 Support to Recompiler 


bayedieng:
Currently Obliteration's recompiler only provides support for x86_64 platforms. In order to support m1 macs and perhaps even Linux ARM/Android devices aarch64 will be required.

yisonPylkita:
After building with
```bash
cmake --preset mac-debug .
cmake --build --preset mac-release
```
I'm stuck on this

```
error[E0570]: `"sysv64"` is not a supported ABI for the current target
  --> kernel/src/process/mod.rs:79:5
   |
79 |     extern "sysv64" fn exit() {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0570]: `"sysv64"` is not a supported ABI for the current target
  --> kernel/src/process/mod.rs:83:5
   |
83 |     extern "sysv64" fn handle_ud2(&mut self, addr: usize) -> ! {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0570`.
error: could not compile `kernel` due to 2 previous errors
make[2]: *** [src/core-prefix/src/core-stamp/core-build] Error 101
make[1]: *** [src/CMakeFiles/core.dir/all] Error 2
make: *** [all] Error 2
```

There are 2 ways around this
1) Implement x86_64 -> aarch64 recompiler
2) Compile this as MacOS Intel binary and let Rosetta 2 do the job

While second option is more of a hack around original problem I wonder if amount of additional work required to provide proper x86 -> aarch64 translation will be worth it. Mind you that `dynarmic` only supports x86_64 -> ARM translations. You would have to either write it yourself or try to use [Fex](https://github.com/FEX-Emu/FEX)/[Box64](https://github.com/ptitSeb/box64)

bayedieng:
We only have support Intel mac atm, this issue is for adding general support to aarch64 devices. While it is possible to use Rosetta 2 for binary translation, we are looking to generally support any aarch64 device that fits the requirements. Dynarmic is an option but it is written in C++, although bindings could be provided it is usually quite difficult to do so with C++.  @ultimaweapon any thoughts?

bayedieng:
[...]

PR #93, opened 2023-01-28
make recompiler target ISA-agnostic

bayedieng:
This PR is the first step towards #73.

### DIFF
--- a/src/kernel/src/process/mod.rs
+++ b/src/kernel/src/process/mod.rs
@@ -76,11 +76,11 @@ impl Process {
         Ok(0)
     }
 
-    extern "sysv64" fn exit() {
+    extern "C" fn exit() {
         // TODO: What should we do here?
     }
 
-    extern "sysv64" fn handle_ud2(&mut self, addr: usize) -> ! {
+    extern "C" fn handle_ud2(&mut self, addr: usize) -> ! {
         info!(
             self.id,
             "Process exited with UD2 instruction from {:#018x}.", addr
--- a/src/kernel/src/process/mod.rs
+++ b/src/kernel/src/process/mod.rs
@@ -76,17 +76,35 @@ impl Process {
         Ok(0)
     }
 
+    #[cfg(target_arch = "x86_64")]
+    extern "sysv64" fn exit() {
+        // TODO: What should we do here?
+    }
+
+    #[cfg(not(target_arch = "x86_64"))]
     extern "C" fn exit() {
         // TODO: What should we do here?
     }
 
+    #[cfg(target_arch = "x86_64")]
+    extern "sysv64" fn handle_ud2(&mut self, addr: usize) -> ! {
+        info!(
+            self.id,
+            "process exited with ud2 instruction from {:#018x}.", addr
+        );
+
+        // fixme: return to "run" without stack unwinding on windows.
+        std::process::exit(0);
+    }
+
+    #[cfg(not(target_arch = "x86_64"))]
     extern "C" fn handle_ud2(&mut self, addr: usize) -> ! {
         info!(
             self.id,
-            "Process exited with UD2 instruction from {:#018x}.", addr
+            "process exited with ud2 instruction from {:#018x}.", addr
         );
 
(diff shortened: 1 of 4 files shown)


src/kernel/src/process/module/recompiler.rs
```diff
@@ -790,7 +803,7 @@ impl<'input> Recompiler<'input> {
     }
 
     fn transform_ud2(&mut self, i: Instruction) -> usize {
-        let handler: extern "sysv64" fn(&mut Process, usize) -> ! = Process::handle_ud2;
+        let handler: extern "C" fn(&mut Process, usize) -> ! = Process::handle_ud2;
```
ultimaweapon, line 806: We also need to fix the code that invoke this on Windows because it uses different calling conventions.

src/kernel/src/process/module/mod.rs
```diff
@@ -194,7 +194,7 @@ pub(super) struct DebugOpts {
     pub original_mapped_dump: PathBuf,
 }
 
-pub(super) type EntryPoint = extern "sysv64" fn(*mut Arg, extern "sysv64" fn());
+pub(super) type EntryPoint = extern "C" fn(*mut Arg, extern "C" fn());
```
ultimaweapon, line 197: Same here. This one is quite complicated because we need to wrap the actual entry point to convert C calling convention to sysv64.

src/kernel/src/process/mod.rs
```diff
@@ -76,11 +76,11 @@ impl Process {
         Ok(0)
     }
 
-    extern "sysv64" fn exit() {
+    extern "C" fn exit() {
```
ultimaweapon, line 79: Not sure if we already invoked this. If yes we need to fix this too.

ultimaweapon (changes requested):

src/kernel/src/process/module/recompiler.rs
```diff
@@ -1,14 +1,16 @@
 use super::Segment;
 use crate::process::Process;
 use iced_x86::code_asm::{
-    byte_ptr, get_gpr64, get_gpr8, qword_ptr, dword_ptr, rax, rdi, rsi, CodeAssembler, CodeLabel,
+    byte_ptr, dword_ptr, get_gpr64, get_gpr8, qword_ptr, rax, rdi, rsi, CodeAssembler, CodeLabel,
```
Mou-Ikkai, line 4: Why'd you move dword_ptr away from qword_ptr?

src/kernel/src/process/module/recompiler.rs
```diff
@@ -169,7 +177,9 @@ impl<'input> Recompiler<'input> {
                 Code::Mov_rm32_r32 => (self.transform_mov_rm32_r32(i), false),
                 Code::Mov_rm64_r64 => (self.transform_mov_rm64_r64(i), false),
                 Code::Movzx_r32_rm8 => (self.transform_movzx_r32_rm8(i), false),
-                Code::Nop_rm16 | Code::Nop_rm32 | Code::Nopd | Code::Nopw => (self.preserve(i), false),
+                Code::Nop_rm16 | Code::Nop_rm32 | Code::Nopd | Code::Nopw => {
+                    (self.preserve(i), false)
+                }
```
Mou-Ikkai, line 182: Unnecessary separation of Nop codes, no longer follows other codes in style.

Mou-Ikkai (commented):
Some trivial styling issues.

ultimaweapon:
@Mou-Ikkai actually we use https://github.com/rust-lang/rustfmt to format code, and it is the reason for @bayedieng changes.

src/kernel/src/process/module/recompiler.rs
bayedieng: Sysv64 is for non-windows x86_64 but it still compiles on windows?

bayedieng: To solve this we could change the sysv64 to x86_64 only and use ```extern C``` for arm as there's no equivalent call convention.

ultimaweapon (approved):

PR #100, opened 2023-02-01
RustFMT addition to GHA

Mou-Ikkai:


ultimaweapon (changes requested):
It is better to run as a step before start building so it is fail fast rather than fail when everything has been built.

.github/workflows/pr.yml
```diff
@@ -5,6 +5,9 @@ on:
     - '*'
 concurrency: pr-${{ github.ref }}
 jobs:
+  quality-check:
+    name: quality-check
```
ultimaweapon, line 9: We should use `QC` or `Quality Check` so its style is matched with the other job.

.github/workflows/pr.yml
```diff
@@ -5,6 +5,9 @@ on:
...
+    uses: ./.github/workflows/QC.yml
```
ultimaweapon, line 10: Better to use lower-case file name when possible because on Linux file name is case-sensitive. Mixed-case file name likely to cause headache on Linux.

.github/workflows/QC.yml
```diff
@@ -0,0 +1,12 @@
...
+  workflow_call: 
+jobs: 
+  qualitycheck: 
+    name: Quality Check
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v3
+    - name: Execute Rustfmt check
+      run: cd src && cargo fmt --all
```
ultimaweapon, line 12: `cargo fmt --all` is used for formatting the source code, which is not what we want here. What we want is `cargo fmt --check` to check if the source code is in a valid style.

ultimaweapon (changes requested):
We need to make other jobs depend on `quality-check` job otherwise our check is useless because it does not prevent the PR from merging if coding style is not correct.

ultimaweapon:
You need to format any source file that cause the QC step to fail. You can do this by right click on anywhere in the source code in VS Code then choose Format Document.

ultimaweapon (approved):

Issue #107, opened 2023-02-02
Rename QC workflow to Pre-build


ultimaweapon:
It just look weird to have `QC / Quality Checks` like:

```
PR Build / QC / Quality Checks (pull_request)
PR Build / Build / Windows (pull_request)
PR Build / Build / Linux (pull_request)
PR Build / Build / Mac (pull_request)
```

VS:

```
PR Build / Pre-build / Quality Checks (pull_request)
PR Build / Build / Windows (pull_request)
PR Build / Build / Linux (pull_request)
PR Build / Build / Mac (pull_request)
```

We can also add other steps that does not related to QC in the future because it not only for QC anymore.

Mou-Ikkai:
Perhaps the QC file will be a general checking file
It just will have multiple jobs in the file for different checks.

ultimaweapon:
Yes that is the idea.

PR #108, opened 2023-02-02
Change QC -> Pre-Build

VocalFan:
Fixes issue #107

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -7,7 +7,7 @@ concurrency: pr-${{ github.ref }}
 jobs:
   quality-checks:
     name: Pre-Build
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -6,7 +6,7 @@ on:
 concurrency: pr-${{ github.ref }}
 jobs:
   quality-checks:
-    name: QC
+    name: Pre-Build
     uses: ./.github/workflows/qualitychecks.yml
```
ultimaweapon, line 10: We should rename this too. Maybe `ci-prebuild.yml`?

ultimaweapon (changes requested):

ultimaweapon (approved):

Issue #106, opened 2023-02-02
Rename obliteration-kernel to obkrnl


ultimaweapon:
Because `obliteration-kernel` is too long and `obkrnl` is not conflict with any *nix programs.

PR #111, opened 2023-02-03
Rename Obliteration-Kernel to obkrnl

VocalFan:
This is just a quick Ctrl-Shift-F job, so tell me if I forgot anything!

Should fix issue #106


ultimaweapon (approved):
It should covered all of it.

PR #116, opened 2023-02-03
Initializes Flatpak builder

### LINKED ISSUE #75: Produce Flatpak package on CI
So Linux users don't need to build from sources.

ultimaweapon:
Closes #75.


.github/workflows/ci-linux.yml
```diff
@@ -8,29 +8,22 @@ jobs:
     steps:
     - name: Checkout source
       uses: actions/checkout@v3
-    - name: Install system dependencies
+    - name: Install System Packages
       run: |
-        wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo tee /etc/apt/trusted.gpg.d/lunarg.asc
```
yisonPylkita, line 13: Do you want to skip Vulkan SDK installation for now?

ultimaweapon: Yeah because we did not use it yet and it seems like we do not need it soon.

yisonPylkita: That's fine. These steps will stay in Git history anyway

ultimaweapon: Right. When we need it we can grab this from the history without required to keep maintenance it while we are not using.

bayedieng (approved):

PR #118, opened 2023-02-04
Linux CI - Caching of Flatpak runtimes.

VocalFan:
Comparing from the latest Dev Build in obhq's builds VS latest build in this PR (I made Linux run twice so it would use the cache)

No cache: 4:50
Flatpak Runtime cache: 4:07

Do note that since the flatpak runtimes are hard set, it'll only invalidate cache if ci-linux.yml is modified.


.github/workflows/ci-linux.yml
```diff
@@ -25,5 +32,5 @@ jobs:
     - name: Upload Flatpak
       uses: actions/upload-artifact@v3
       with:
-        name: obliteration-linux
+        name: obliteration-linux64
```
ultimaweapon, line 35: Now we are going to support both AMD64 and ARM64 so this name is ambiguous. We should rename it to something like `linux-x86_64` for Linux and `windows-x64` for Windows.

ultimaweapon (changes requested):

VocalFan:
I made it be:

Intel/AMD: x64
Arm: arm64

Just to make it short and simple.

.github/workflows/ci-linux.yml
```diff
@@ -32,5 +32,5 @@ jobs:
     - name: Upload Flatpak
       uses: actions/upload-artifact@v3
       with:
-        name: obliteration-linux64
+        name: obliteration-linux-x64
```
ultimaweapon, line 35: Better to use `x86_64`, `x86-64` or `amd64` because x64 is not an official architecture name on Linux world.

ultimaweapon (changes requested):

.github/workflows/ci-linux.yml
VocalFan: This is a hate crime.

ultimaweapon (approved):

PR #120, opened 2023-02-04
Compile on aarch64

yisonPylkita:
With this Obliteration will compile on Apple M1. 

I'll try to setup aarch64 CI

### DIFF
--- a/src/kernel/src/process/module/recompiler/aarch64.rs
+++ b/src/kernel/src/process/module/recompiler/aarch64.rs
@@ -1,0 +1,484 @@
+use super::{LabelType, NativeCode, Recompiler, RunError};
+use crate::process::module::Segment;
+use crate::process::Process;
+use iced_x86::code_asm::{CodeAssembler, CodeLabel};
+use iced_x86::{Instruction, Register};
+use std::collections::{HashMap, VecDeque};
+
+pub struct Aarch64Emitter<'input> {
+    proc: *mut Process,
+    input: &'input [u8],
+    segments: Vec<Segment>,
+    assembler: CodeAssembler,
+    jobs: VecDeque<(u64, usize, CodeLabel)>,
+    labels: HashMap<u64, CodeLabel>,
+    output_size: usize,
+}
+
+impl<'input> Aarch64Emitter<'input> {
+    pub fn new(proc: *mut Process, input: &'input [u8], segments: Vec<Segment>) -> Self {
+        Self {
+            proc,
+            input,
+            segments,
+            assembler: CodeAssembler::new(64).unwrap(),
+            jobs: VecDeque::new(),
+            labels: HashMap::new(),
+            output_size: 0,
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


ultimaweapon:
Which CI you are going to use? AFAIK Apple M1 is not available on GitHub Actions yet.

yisonPylkita:
I don't know yet. GitHub doesn't support arm for cloud yet (you can run on self-hosted runner though). Travis, CircleCI, Oracle CI, and others seems to require credit cards. 

I have a Raspberry Pi 4 laying around but I it could be super slow. I'll come back with a different PR when I have something. For now lets mere this if you don't have objections

src/kernel/src/process/module/recompiler/aarch64.rs
```diff
@@ -0,0 +1,484 @@
...
+use iced_x86::{Instruction, Register};
+use std::collections::{HashMap, VecDeque};
+
+pub struct Aarch64Emitter<'input> {
+    proc: *mut Process,
+    input: &'input [u8],
+    segments: Vec<Segment>,
+    assembler: CodeAssembler,
+    jobs: VecDeque<(u64, usize, CodeLabel)>,
+    labels: HashMap<u64, CodeLabel>,
```
bayedieng, line 14: These fields should be removed as they're specific to x86_64 instructions. We won't be using the same assembler.

src/kernel/src/process/module/recompiler/aarch64.rs
```diff
@@ -0,0 +1,484 @@
...
+
+impl<'input> Aarch64Emitter<'input> {
+    pub fn new(proc: *mut Process, input: &'input [u8], segments: Vec<Segment>) -> Self {
+        Self {
+            proc,
+            input,
+            segments,
+            assembler: CodeAssembler::new(64).unwrap(),
+            jobs: VecDeque::new(),
+            labels: HashMap::new(),
```
bayedieng, line 26: To be removed as well

ultimaweapon:
It should be possible to cross compile for M1 from Intel, I guess. Let @bayedieng review this PR.

src/kernel/src/process/module/recompiler/aarch64.rs
```diff
@@ -0,0 +1,484 @@
...
+            output_size: 0,
+        }
+    }
+
+    fn temp_register64(keep: Register) -> Register {
+        unimplemented!();
+    }
+}
```
bayedieng, line 34: Also to be removed.

bayedieng (changes requested):

bayedieng:
> It should be possible to cross compile for M1 from Intel, I guess. Let @bayedieng review this PR.

Yes it is possible to do so by running ```rustup target add x86_64-apple-darwin``` in the CLI. It should be possible to run a game in theory using Rosetta 2 as well but have not tried this.

bayedieng (approved):

PR #122, opened 2023-02-04
Mechanism to build DEB packages

yisonPylkita:
I know there is already a mechanism to generate Flatpaks but I've been trying to generate MacOS bundle via CPack and nothing works. Yet generating DEB packages works fine. I think I'm loosing my sanity 😅

If you think this mechanism is redundant then please feel free to close this PR.

PS: I hate CMake!


ultimaweapon:
I'm not sure if one `deb` package can work across multiple distro and version (e.g. work on both Debian and Ubuntu 20.04 22.04 etc.). If not maybe we should stick with Flatpak or something similar.

ultimaweapon:
Regarding the discussion in #123 maybe we should focus on AppImage instead of distro-specific package?

PR #125, opened 2023-02-05
Refactors exFAT

ultimaweapon:
After I was refactored the PFS library I found out the pattern that I used with it is more flexible on the user side because the user don't need to keep the root object while working with its file. So I did the same to exFAT.


src/pup/src/lib.rs
```diff
@@ -106,11 +106,14 @@ impl Pup {
...
         // Create exFAT reader.
-        let _fat = match ExFat::open(entry) {
+        let fat = match ExFat::open(entry) {
             Ok(v) => v,
             Err(e) => return Err(DumpSystemImageError::CreateImageReaderFailed(e)),
         };
 
+        // Dump files.
+        for item in fat {}
+
```
bayedieng, line 116: Was there supposed to be something implemented in this loop?

bayedieng (changes requested):

src/pup/src/lib.rs
ultimaweapon: Right. We need a complete implementation of `Directory` and `File` from `exfat` before we can implement this loop.

bayedieng (approved):

PR #127, opened 2023-02-06
Add checks for Rust-Clippy

VocalFan:


.github/workflows/qc-prebuild.yml
```diff
@@ -10,3 +10,11 @@ jobs:
...
     - name: Run RustFMT Check
       run: cd src && cargo fmt --check
+  rust-clippy:
+    name: Rust-Clippy Check
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v3
+    - name: Run Rust-Clippy Check
+      run: cd src && cargo clippy
```
ultimaweapon, line 20: Better to merge it with a rustfmt job as a step so we won't have too many jobs to show in the PR.

ultimaweapon (changes requested):

.github/workflows/qc-prebuild.yml
VocalFan: Ah, no multi-thread job? But then doesn't that make 
`Pre-Build / Rustfmt Check`
Be untrue since we use multiple commands?

ultimaweapon: The parallel is not necessary for this case because it not taking much time for each step.

src/pfs/src/file.rs
```diff
@@ -167,6 +167,7 @@ impl<'a> Seek for File<'a> {
 }
 
 impl<'a> Read for File<'a> {
+    #[allow(clippy::uninit_vec)]
```
ultimaweapon, line 170: Could you wrap the problematic lines inside `{}` then apply this to that block instead?

src/pfs/src/pfsc.rs
```diff
@@ -75,6 +75,7 @@ impl<F: Read + Seek> Reader<F> {
         self.original_size
     }
 
+    #[allow(clippy::uninit_vec)]
```
ultimaweapon, line 78: Same here.

src/util/src/mem/mod.rs
```diff
@@ -4,10 +4,12 @@ use std::mem::{size_of, MaybeUninit};
...
 
 /// Just a shortcut to `MaybeUninit::uninit().assume_init()`.
-pub fn uninit<T>() -> T {
+#[allow(clippy::uninit_assumed_init)]
+pub unsafe fn uninit<T>() -> T {
     unsafe { MaybeUninit::uninit().assume_init() }
 }
 
+#[allow(clippy::uninit_vec)]
 pub fn new_buffer<T>(size: usize) -> Vec<T>
```
ultimaweapon, line 13: We should mark this one as unsafe too because it can cause undefined behavior if not using correctly.

ultimaweapon (approved):

Issue #13, opened 2022-10-07
Handle ANSI escape code in kernel log viewer


ultimaweapon:
So we can print a beautiful text from the kernel :). We can use this code from Qt Creator. Here is the following source that handled ANSI escape on Qt Creator:

[ansiescapecodehandler.h](https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/libs/utils/ansiescapecodehandler.h)
[ansiescapecodehandler.cpp](https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/libs/utils/ansiescapecodehandler.cpp)
[outputformatter.cpp](https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/libs/utils/outputformatter.cpp)

PR #131, opened 2023-02-06
Initializes ANSI escape code supports

ultimaweapon:
First step of #13. Also update the README about unsafe rule and fix another possible GUI crash when the kernel crash.

Screenshot:
Screenshot from 2023-02-06 22-41-47


ultimaweapon:
@bayedieng wait, I saw a memory leak.

bayedieng (approved):

PR #132, opened 2023-02-07
Implement AVX registers

VocalFan:
I get it, it looks horrible, I tried to format it the best I could. 
 Stillllllllll waiting on Ultima to do his side, so I can only go this far.


src/kernel/src/process/module/recompiler/x64.rs
```diff
@@ -1757,9 +1911,29 @@ impl Recompiler for X64Emitter<'_> {
...
+            let src = i.ip_rel_memory_address();
+            let tmp = get_gpr64(Self::temp_register64(dst)).unwrap();
+
+            if self.is_executable(src) {
+                // Call function to execute code at source memory location
+                self.assembler.call(src).unwrap();
+                // Move the result to the destination register
+                self.assembler
+                    .vmovups(get_ymm(dst).unwrap(), ymmword_ptr(tmp))
+                    .unwrap();
```
ultimaweapon, line 1924: `vmovups` is using for copy data so this block is not correct.

src/kernel/src/process/module/recompiler/x64.rs
```diff
@@ -1757,9 +1911,29 @@ impl Recompiler for X64Emitter<'_> {
     fn transform_vmovups_ymm_ymmm256(&mut self, i: Instruction) -> usize {
         // Check if second operand use RIP-relative.
         if i.op1_kind() == OpKind::Memory && i.is_ip_rel_memory_operand() {
-            panic!(
-                "VMOVUPS ymm1, ymm2/m256 with second operand as RIP-relative is not supported yet."
-            );
+            let dst = i.op0_register();
+            let src = i.ip_rel_memory_address();
+            let tmp = get_gpr64(Self::temp_register64(dst)).unwrap();
```
ultimaweapon, line 1916: We don't need this in this case because the destination operand is `ymm`, which mean we can use any general purpose register (e.g. `rax`) without conflicting.

src/kernel/src/process/module/recompiler/x64.rs
```diff
@@ -52,6 +52,111 @@ impl<'input> X64Emitter<'input> {
...
+            Register::XMM23 => Register::XMM22,
+            Register::XMM24 => Register::XMM25,
+            Register::XMM25 => Register::XMM24,
+            Register::XMM26 => Register::XMM27,
+            Register::XMM27 => Register::XMM26,
+            Register::XMM28 => Register::XMM29,
+            Register::XMM29 => Register::XMM28,
+            Register::XMM30 => Register::XMM31,
+            Register::XMM31 => Register::XMM30,
+
```
ultimaweapon, line 159: With the reason from the previous comment mean we don't need this.

ultimaweapon (changes requested):

PR #152, opened 2023-02-21
Replaces PUP with system files from PS4

ultimaweapon:
We found out the SPRXs we got from the update file is encrypted and currently no way to decrypt it without installing it on a jailbroken PS4 and pull it back using a specialized FTP server. So we drop the PUP supports.


README.md
```diff
@@ -23,21 +23,15 @@ We have a Matrix Room Space `#obliteration:matrix.org` on [Matrix.to](https://ma
 
 ## Features
 
-- [x] Built-in PUP file supports for decrypted PUP from [pup_decrypt](https://github.com/idc/ps4-pup_decrypt).
+- [x] Built-in FTP client to pull the decrypted firmware from jailbroken PS4.
 - [x] Built-in PKG file supports for Fake PKG.
 - [x] Game library.
 - [ ] Emulate system calls instead of user-space libraries.
 
 ## System requirements
```
bayedieng, line 31: Add somewhere in the requirements that a jailbroken ps4 is necessary.

src/ftp/src/lib.rs
```diff
@@ -0,0 +1,520 @@
+use self::retrieve::Retrieve;
+use std::fmt::Display;
+use std::io::{BufRead, BufReader, Write};
+use std::net::TcpStream;
+use thiserror::Error;
+
+pub mod retrieve;
+
+/// The reason we need to implement our own FTP client is because all of other FTP client is not
+/// support custom command.
```
bayedieng, line 10: slight grammer issue here. Should be "The reason we need to implement our own FTP client is because no other FTP client supports the ```custom``` command".

bayedieng (changes requested):

README.md
bayedieng: should include that an ftp server should be installed as well.

ultimaweapon: Just added it to the system requirements section.

bayedieng (approved):

VocalFan (approved):

bayedieng (changes requested):
There are some cases where the user will already have a decrypted sprx files available, there should be in option in the wizard to point to them directly

bayedieng (approved):

PR #157, opened 2023-02-25
Adds LLVM

ultimaweapon:


ultimaweapon:
Our current problem: How to make `cc` crate use the same MSVC version as CMake?

src/kernel/src/llvm/module.rs
```diff
@@ -0,0 +1,71 @@
...
+use std::ffi::c_char;
+use std::ptr::null_mut;
+use std::sync::{Mutex, MutexGuard};
+
+/// A wrapper on LLVM module for thread-safe.
+pub struct LlvmModule<'a> {
+    context: &'a Mutex<LLVMContextRef>,
+    raw: LLVMModuleRef,
+    _lock: MutexGuard<'a, LLVMContextRef>,
+}
```
bayedieng, line 16: should have more consistency with LLVM name. Should be ```LLVMModule```.

src/kernel/src/llvm/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl Drop for Llvm {
+    fn drop(&mut self) {
+        unsafe { LLVMContextDispose(*self.context.get_mut().unwrap()) };
+    }
+}
+
+/// A wrapper on LLVM context for thread-safe.
+pub struct LLvmContext<'a> {
```
bayedieng, line 40: should be ```LLVMContext```

bayedieng (changes requested):
Overall should have more consistent naming ```LLVM``` should always be all capitalized

src/kernel/src/llvm/module.rs
ultimaweapon: The problem is that one is not consistent with Rust naming. If you take a look on Rust `std` you will see they always use pascal-case on an abbreviation (e.g. `TypeId` instead of `TypeID`).

bayedieng (approved):

PR #158, opened 2023-02-26
turn off zstd on arm64 llvm build

bayedieng:


src/CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ else()
         list(APPEND LLVM_OPTS -DLLVM_TARGETS_TO_BUILD:STRING=X86)
     elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "arm64")
         list(APPEND LLVM_OPTS -DLLVM_TARGETS_TO_BUILD:STRING=AArch64)
+
+        # arm64 LLVM default requires lzstd to be built dynamically.
+        list(APPEND LLVM_OPTS -DLLVM_ENABLE_ZSTD:BOOL=OFF)
```
ultimaweapon, line 24: Better to turn it off for every platform because we don't use it and it is likely to cause the same problem on every platform.

ultimaweapon (changes requested):

ultimaweapon:
Sorry I mean turn off for every platforms, including Windows 😂

bayedieng:
> Sorry I mean turn off for every platforms, including Windows joy

didn't realize windows and linux build were different 😅.

ultimaweapon (approved):

PR #216, opened 2023-06-02
Add support for Dark Mode (And move to Fusion)

VocalFan:
This modifies the theme to Fusion for Dark Mode support on Windows! And for the Wizard, we now use the style Modern due to Aero breaking window dragging and theming. Along with that, dark mode on all OSes should have white icons now! *Very cool, of course :P*


ultimaweapon:
We should wait until the new setup has been merged because this will break setup screen.

src/initialize_wizard.cpp
```diff
@@ -274,6 +274,7 @@ InitializeWizard::InitializeWizard()
 {
     // Window properties.
     setWindowTitle("Setup Obliteration.");
+    setWizardStyle(QWizard::ModernStyle);
```
ultimaweapon, line 277: We should explicitly set this on Windows only so on the other platform it will still looks native.

src/main.cpp
```diff
@@ -13,6 +13,7 @@ int main(int argc, char *argv[])
     // Setup application.
     QCoreApplication::setOrganizationName("Obliteration");
     QCoreApplication::setApplicationName("Obliteration");
+    QApplication::setStyle("Fusion");
```
ultimaweapon, line 16: Same here.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #230, opened 2023-06-14
Added SimpleLog implementation

VocalFan:
This removes our *I'll say extremely restrictive* custom logging system and instead uses SimpleLog, allowing for a more detailed view of what's going on. Along with this, a log file will be written to when the kernel is executed in the root folder of the application, named `obliteration-kernel.log` *The filename can be changed if wanted, just a quick name for it.*


VocalFan:
Hopefully now that I used the ProjectDir, you'll use that for your config instead of a hardset Registry key lol.

(Also wow, RustFMT just made it even less consistent in style :<)

ultimaweapon:
I don't agree with using a separated mechanism to get the path to write the log than the one that provided from the main application. This will cause a multiple source of truth which end up more than one locations for the debugging data.

The log formatting also is hard to read because the category and the message is too far from each other. The printing of `std::error::Error` object is also unclear if it will printing the nested error, which is important for me to identity the root of problem.

VocalFan:
> I don't agree with using a separated mechanism to get the path to write the log than the one that provided from the main application. This will cause a multiple source of truth which end up more than one locations for the debugging data.
> 
> The log formatting also is hard to read because the category and the message is too far from each other. The printing of `std::error::Error` object is also unclear if it will printing the nested error, which is important for me to identity the root of problem.

It does print the nested error. It's the same as the implementation before, except that it's adopted to fit the type of String that the Log crate wants. Also, for the 1st concern, it's a bit more trustworthy imo to use native Rust instead of Qt C++ for getting the path. Makes it not have to depend on the framework to get the path.

ultimaweapon:
Given the following code:

```rs
use log::error;
use simplelog::*;
use thiserror::Error;

fn main() {
    TermLogger::init(
        LevelFilter::Info,
        Config::default(),
        TerminalMode::Mixed,
        ColorChoice::Auto
    );

    let e = FooError::Error1(BarError::Error2);

    error!("Something wrong: {}", e);
}

#[derive(Debug, Error)]
enum FooError {
    #[error("error 1")]
    Error1(#[source] BarError),
}

#[derive(Debug, Error)]
enum BarError {
    #[error("error 2")]
    Error2,
}
```

Its output is:

```
06:02:43 [ERROR] Something wrong: error 1
```

Which show it does not handled the nested error because it does not print both `error 1` and `error 2`.

The main problem here with the path is a multiple source of truth.

VocalFan:
@ultimaweapon Fixed. This is simply done by using {:#} instead of {}.

VocalFan:
I've also reduced the crowdedness a bit, check it out! :P

ultimaweapon:
Your `{:#}` is not working:

```rs
use log::error;
use simplelog::*;
use thiserror::Error;

fn main() {
    TermLogger::init(
        LevelFilter::Info,
        Config::default(),
        TerminalMode::Mixed,
        ColorChoice::Auto
    );

    let e = FooError::Error1(BarError::Error2);

    error!("Something wrong: {:#}", e);
}

#[derive(Debug, Error)]
enum FooError {
    #[error("error 1")]
    Error1(#[source] BarError),
}

#[derive(Debug, Error)]
enum BarError {
    #[error("error 2")]
    Error2,
}
```

Output:

```
05:54:45 [ERROR] Something wrong: error 1
```

VocalFan:
@ultimaweapon Check Element, you get a choice for which one, since I tested and, you were right, but there's two formats I can use for nested errors.

ultimaweapon:
I'll not participate in the real-time chat regarding to the development of Obliteration anymore. Please provide the formats that you mentioned here.

VocalFan:
> I'll not participate in the real-time chat regarding to the development of Obliteration anymore. Please provide the formats that you mentioned here.

 Oh... Sorry... .m. Ahem... So yes, {:#} was wrong, but that's simply because I was missing a symbol.
 
 
image
This is using {:?} ^

image
And this is using {:#?} ^
Basically, pretty-prints the nest.

VocalFan:
Realized the snipping tool failed to save to clipboard on that 2nd pic... Fixed! And still, sorry, Ultima...

ultimaweapon:
Both of them printing the enum identifier, not the message associated with the enum. It should output `error 1` and `error 2` instead of `Error1` and `Error2`. It seems like this cannot be done with the current version of `log` crate. So we may need to postpone this until a solution is available.

VocalFan:
> Both of them printing the enum identifier, not the message associated with the enum. It should output `error 1` and `error 2` instead of `Error1` and `Error2`. It seems like this cannot be done with the current version of `log` crate. So we may need to postpone this until a solution is available.

I FIGURED IT OUT! :3
```rs
use log::error;
use simplelog::*;
use thiserror::Error;

fn main() {
    TermLogger::init(
        LevelFilter::Info,
        Config::default(),
        TerminalMode::Mixed,
        ColorChoice::Auto
    );

    let e = FooError::Error1(BarError::Error2);

    error!("Something wrong: {}", e);
}

#[derive(Debug, Error)]
enum FooError {
    #[error("Bad stuff {0}")]
    Error1(#[source] BarError),
}

#[derive(Debug, Error)]
enum BarError {
    #[error("And bad things")]
    Error2,
}
```

Using {0} at the end, we can see the nested error.
image

ultimaweapon:
That is not a good one because it required to modify the error type, which is a problem if the error coming from the external crates.

VocalFan:
But... It is, Ultima. Because the crate would already have it included if needed, and excluded if not needed. The errors are handled by the crate themselves... They implement their log code.

ultimaweapon:
That is not how Rust error handling is working.

VocalFan:
> That is not a good one because it required to modify the error type, which is a problem if the error coming from the external crates.

Yeah uh, me and the entire Rust discord server have no clue what you mean .w.

```
{0} would work if you want to print the inner error. not sure what your friend means tbh
```

ultimaweapon:
If my above example is an external crate you will never be able to print the inner one with your solution.

VocalFan:
...*Still, everyone disagrees with you.*

VocalFan:
I don't mean it to be rude, but I'm just saying... Everyone in the Rust server is confused on what you're saying because... It isn't going to be an issue.

ultimaweapon:
Every developer has different experience like I said before. Those developers think it is not an issue but it is an issue for me. They are not working on this project so what is the point of following their advise that I believed it is going to cause some issues for me who are working on this project? Please don't forget this fact.

VocalFan:
Well, I guess I just wait then.

VocalFan:
Shall I turn this into a draft? 

*And still, sorry for what happened in Element... Sucks not being able to chat with ya... Oh, plus someone had a question in General!*

ultimaweapon:
If you believed you can fix all of the issues soon then leave this PR opened or make it draft otherwise close it and rework when a solution is available will be better so you don't need to keep merging the main branch into this, which is likely to encounter some conflicts that you need to resolve from time to time.

I found out that leaving from the real-time chat is better for me to focus on working so I'll leave our real-time chat to your responsibility or if you don't want to handle this feel free to close it. If you want to keep it maybe switching to Discord should be better. Please create a PR to update the README if you choose to close it or switch to Discord.

VocalFan:
# Drafted until solution is found for Nested errors.

ultimaweapon:
I don't think we need to use the external logging anymore so I'll close this.

PR #252, opened 2023-07-13
Added/Renamed Errors.

VocalFan:
Implemented all Kernel, SRTC, DevKit, NPDRM, Auth Manager, System File, Safemode, and USBD based errors into errno.rs along with changing all files that depended on the old name.

*The reason that all kernel based errors start with KERNEL_ is that, besides from starting with `SCE_KERNEL_ERROR_`, I also did this since other segments have overlapping names. Such as: `KERNEL_ENOSPC` and `NPDRM_ENOSPC`*


VocalFan:
...Welp, I made our Qt's aqtinstall version newer, but even that didn't fix the issue... Seems it wants to install Qt 6.6.0 but cant.

ultimaweapon:
How did you get those values from? I don't think the `KERNEL_` prefix is correct because the FreeBSD kernel is also using the plain name, which I don't believe Sony is going to rename it.

VocalFan:
> How did you get those values from? I don't think the `KERNEL_` prefix is correct because the FreeBSD kernel is also using the plain name, which I don't believe Sony is going to rename it.
<html>
<body>


SCE_KERNEL_ERROR_EPERM | CE-30001-4 | 0x80020001 | Operation not permitted / An error has occurred (error with the application)
-- | -- | -- | --
SCE_KERNEL_ERROR_ENOENT | CE-30002-5 | 0x80020002 | Application data may be corrupted


</body>
</html>

Besides from the errors starting with SCE_KERNEL_ERROR_? I also did it due to overlapping errors that'd be an issue if we just used like... `enoent enoent enoent` as there's one for SCE_KERNEL, SCE_DEVACT, and SCE_NET_ERROR.

ultimaweapon:
It seems like those values are for the user-mode libraries, not the kernel. In this case it cannot be used with Obliteration because we emulate the kernel, not the user-mode libraries.

VocalFan:
> It seems like those values are for the user-mode libraries, not the kernel. In this case it cannot be used with Obliteration because we emulate the kernel, not the user-mode libraries.

That makes more sense... Alright, give me a min... As I assume all those other errors aren't exactly needed.

ultimaweapon:
You need to verify those values with the PS4 kernel to make sure it is exactly the same. To verify the value, find the same function in both FreeBSD and PS4 kernel that is using the value you want to verify. Then looking at the location where the errno is returned in the FreeBSD kernel. Compare that location on the PS4 kernel to see if the returned value is the same.

VocalFan:
From what I looked at, they do seem to match. This is, of course, using FreeBSD 9.1's kernel as that's what ORBIS uses.

ultimaweapon:
Yeah so far I did not see any of it that is different. But I don't want to have an headache when something is not working as expected because of the wrong errno. So better to make sure each value is actually matched.

VocalFan:
Well that's what I mean, I checked using FreeBSD 9.1 lol

ultimaweapon:
The PS4 kernel may use a different value than FreeBSD. Although it is unlikely but better to make sure each value is actually matched because the reason that I was stated above.

VocalFan:
...Okay, we seem to be on the wrong foot.

I mean

I checked against the PS4 kernel, and the FreeBSD 9.1 kernel. My lad.

ultimaweapon:
How did you check on the PS4 kernel?

VocalFan:
> How did you check on the PS4 kernel?



> How did you check on the PS4 kernel?

Well, as you said, I tried to find functions from both kernels! Though, you know me not always getting things right first try... If you wanna give me a guide to double check off, that'd be neat... But from all the ones I tested... Such as EINVAL, ESRCH, EDEADLK... They all seemed to line up. Still, would be best to have a proper set of steps to double check, especially since you like things being perfectly tested x3

ultimaweapon:
Okay I'll give you a guide for this maybe tomorrow. Today I'm already exhausted.

VocalFan:
> Okay I'll give you a guide for this maybe tomorrow. Today I'm already exhausted.

Just don't die, my man, you're the leader here... Plus you're cool lol.

ultimaweapon:
The easiest way should be compare the syscall functions. First you need to create a new structure in Data Type Manager in Ghidra. The structure you need to create is [sysent](https://github.com/freebsd/freebsd-src/blob/release/9.1.0/sys/sys/sysent.h#L56) and make sure the final size of the structure is 0x30. Then go to address `0xffffffffa1ce8310` and set data type to `sysent[678]`. Then you can compare each `sy_call` with the [FreeBSD](https://github.com/freebsd/freebsd-src/blob/release/9.1.0/sys/kern/init_sysent.c#L36) version.

VocalFan:
image
Don't think I did it right...

VocalFan:
@ultimaweapon 

image

VocalFan:
One merge of main later, and I'd like to say that it seems the SY_CALLs line up! Still find it odd I had to hit the Pack button for Ghidra to properly size the values...

ultimaweapon:
Keep the values that you have verified it is exactly matched. Remove all of them that you can't find where it is used in the PS4.

VocalFan:
Done!

ultimaweapon:
How did you verified all of it? It is quite hard for me to verify all of those values.

VocalFan:
> How did you verified all of it? It is quite hard for me to verify all of those values.

Went off of the values used and the functions Ghidra gave for the sy_calls, compared against the URL to the FREEBSD calls you gave me.

ultimaweapon:
Can you provides more detail on that? Analyzing the code on Ghidra is time consuming and all I got from syscall 0 to 4 is just `ENOSYS`.

VocalFan:
Sooooo, do ya believe me yet? Seeing as all the Error Numbers you've added after my PR seem to match up :P

ultimaweapon:
No. Until you show the evidence that you have verified all of the values you have added.

VocalFan:
Hell, I can even compare it against the open-source version of Sony's Error Code lookup. https://github.com/xXxTheDarkprogramerxXx/PS4_Error_Code_Viewer

ultimaweapon:
PS dev wiki can be used as a reference but cannot be trusted. Where is the errno in that project? All I can find is the Sony error codes.

VocalFan:
> It seems like those values are for the user-mode libraries, not the kernel. In this case it cannot be used with Obliteration because we emulate the kernel, not the user-mode libraries.

Well, you say that and it's... Half true. All of those aren't usable except SCE_KERNEL... PS4 modding and homebrew servers agree with that, and all the SCE_KERNEL error codes have the same value as the PS4 kernel error codes, just that they use a hex at the start, like for EPERM: 0x80020001 instead of just 1. They all just use a 0x8002 starting Hex.

ultimaweapon:
Someone agree with you does not matter unless they has proved your added errno are the same as PS4 kernel.

VocalFan:
> Someone agree with you does not matter unless they has proved your added errno are the same as PS4 kernel.

Someone disagreeing with you does not matter unless they have proved your added errno are not the same as the PS4 kernel.

The CE- codes use the exact same numbers as the FreeBSD codes up until the new ones 95 and higher.

Plus you'd think Sony would have the CE- codes use the same as the FreeBSD codes as they're the same kernel, with the same error messages, same error numbers, and same error names.

VocalFan:
For which @Al-Azif says about 95-99
image

VocalFan:
For your first question, I was just reversing what you gave me, innocent until proven guilty, not guilty until proven innocent.

And I was asking Al-Azif about the error numbers for the kernel, sent him our errno.rs code, and told him we were doing the PS4 Kernel error numbers.

VocalFan:
From LM *The creator of the homebrew menu I use for dumping my disks*
And DarkProgrammer's error code lists.

ultimaweapon:
Those are `SCE_KERNEL_ERROR_*` that you assume it is the same as errno, is not it?

VocalFan:
> Those are `SCE_KERNEL_ERROR_*` that you assume it is the same as errno, is not it?

Eh, kinda? The latest FreeBSD kernel also has Integrity Check Failed, so it's also most likely Sony added their own Integrity Check. *I mean, Integrity Checks are quite common for playstation consoles...*

VocalFan:
But if you wish, we can just keep all the FreeBSD kernel errnos for now, and add those back in if we need them,

VocalFan:
And done!

.github/workflows/ci-linux.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       uses: jurplel/install-qt-action@v3
       with:
         version: 6.*
+        aqtversion: '==3.1.*'
```
ultimaweapon, line 28: What is this?

.github/workflows/ci-mac.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       uses: jurplel/install-qt-action@v3
       with:
         version: 6.*
+        aqtversion: '==3.1.*'
```
ultimaweapon, line 28: Same here.

.github/workflows/ci-windows.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       uses: jurplel/install-qt-action@v3
       with:
         version: 6.*
+        aqtversion: '==3.1.*'
```
ultimaweapon, line 29: Same here.

ultimaweapon (approved):

PR #258, opened 2023-07-14
Added File Logging

VocalFan:
Is this probably janky? ...Yeah.

But am I working off our Logging system due to the Rust Log system not working properly with nested errors.

But yeah, this adds file logging!


src/kernel/src/main.rs
```diff
@@ -27,7 +28,13 @@ mod sysctl;
 
 fn main() -> ExitCode {
     // Initialize logger.
-    let logger = Logger::new();
+    let project_dirs = ProjectDirs::from("", "OBHQ", "Obliteration").unwrap();
```
ultimaweapon, line 31: See my first paragraph at https://github.com/obhq/obliteration/pull/230#issuecomment-1594040399

ultimaweapon (changes requested):

src/kernel/src/main.rs
VocalFan: Which, once again, as I said... It'd be better to use a path from Rust instead of Qt. *Especially since you wanna try using Slint.*

ultimaweapon: It does not matter even with Slint. It still need to use the path that was supplied from the GUI.

VocalFan: Then can you show me where the path from the GUI is? I'll go off that, then.

ultimaweapon: You need to set the log file to `Logger` after the debug dump has been cleared otherwise it is going to remove the log file for the current session. Keep the logger initialized as the first thing like it currently did (without log file) then adding a new method to `Logger` to set the file. Use `args.debug_dump` for a directory to keep the file.

VocalFan:
There :3

ultimaweapon (approved):
There are some minor issues but overall is looking good. I'll fix the remaining issues later.

PR #261, opened 2023-07-20
Fix build issue for arm macs

### LINKED ISSUE #260: Does Not Build on MacOS ARM
the following error is provided when attempting to build on an arm mac, even when cross-compiling to x64 

```
error[E0570]: `"aapcs"` is not a supported ABI for the current target
  --> kernel/src/syscalls/mod.rs:47:5
   |
47 |     pub unsafe fn invoke(&self, i: &Input, o: &mut Output) -> i6...
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0570]: `"aapcs"` is not a supported ABI for the current target
  --> kernel/src/syscalls/mod.rs:88:5
   |
88 |     pub unsafe fn int44(&self, offset: usize, module: &VPathBuf) -> ...
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ```
  Seems to be an issue with the ```cpu_abi``` proc macro.

bayedieng:
This PR fixes building on arm macs by switching to the ```extern "C"``` abi and fixing an erroneous macro input for the error macro. Closes #260

### DIFF
--- a/src/kernel-macros/src/cpu_abi.rs
+++ b/src/kernel-macros/src/cpu_abi.rs
@@ -21,7 +21,7 @@ pub fn transform(item: ItemFn) -> syn::Result<TokenStream> {
     #[cfg(target_arch = "x86_64")]
     let abi = "sysv64";
     #[cfg(target_arch = "aarch64")]
-    let abi = "aapcs";
+    let abi = "C";
     let attrs = item.attrs;
     let vis = item.vis;
     let safety = item.sig.unsafety;
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -216,7 +216,10 @@ fn main() -> ExitCode {
         }
         #[cfg(not(target_arch = "x86_64"))]
         ExecutionEngine::Native => {
-            error!("Native execution engine cannot be used on your machine.");
+            error!(
+                logger,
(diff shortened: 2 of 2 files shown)


ultimaweapon:
The problem with `C` is it platform-specific (e.g. macOS vs Linux). Have you tried the other values?

bayedieng:
Yes unfortunately extern "C" seems to be the only abi compatible with arm64 

Screenshot 2023-07-20 at 1 37 23 PM

aapcs seems to be 32-bit arm only

ultimaweapon:
Strange. Why `aapcs` is not working on macOS.

bayedieng:
I believe it might be 32-bit only, there is an experimental ABI in the works though for 64-bit ARM I believe https://github.com/rust-lang/rust/issues/65815

bayedieng:
There should be an aapcs64 available but I haven't found the ABI.

ultimaweapon:
Seems like our only solution is fixing this on the call sites.

ultimaweapon (approved):

Issue #265, opened 2023-07-21
Unimplemented Syscall 608


VocalFan:
`thread 'main' panicked at 'not yet implemented: syscall 608 at 0x0000000000001eba on /system/common/lib/libkernel.sprx', kernel\src\syscalls\mod.rs:68:18`

CUSTOM PS4 SYSCALL

<html>
<body>

608 | <=1.01? | sys_dynlib_get_info_ex | int sys_dynlib_get_info_ex(int moduleHandle, struct Unk *unk, int *destModuleInfoEx)
-- | -- | -- | --

</body>
</html>

VocalFan:
@ultimaweapon Latest PR gives `Syscall 608 failed: invalid argument.`

Just letting ya know when you do part 2.

VocalFan:
@ultimaweapon 

```
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000005ed0:kernel\src/main.rs:115
Starting Obliteration Kernel.
System directory    : G:\obSTORAGE\dev_sys
Game directory      : G:\obSTORAGE\dev_hdd0\CUSA04518
Debug dump directory: C:\Users\Windows Games\AppData\Local\OBHQ\Obliteration\kernel
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000005ed0:kernel\src/main.rs:133
Initializing file system.
++++++++++++++++++ I [00:00:00:00:002]:0x0000000000005ed0:kernel\src/main.rs:144
Initializing memory manager.
++++++++++++++++++ I [00:00:00:00:002]:0x0000000000005ed0:kernel\src/main.rs:147
Page size is             : 4096
Allocation granularity is: 65536
++++++++++++++++++ I [00:00:00:00:002]:0x0000000000005ed0:kernel\src/main.rs:160
Initializing virtual process.
++++++++++++++++++ P [00:00:00:00:003]:0x0000000000005ed0:G:\GitHub\obliteration\src\tls\src\lib.rs:60
called `Result::unwrap()` on an `Err` value: Os { code: 6, kind: Uncategorized, message: "The handle is invalid." }
```

VocalFan:
By that reaction... Is that unexpected?

VocalFan:
Oh! If I do it in the Kernel using VSCode:

```
; Symbol: __rust_start_panic
; Source: \rustc\8ede3aae28fe6e4d52b38157d7bfe0d3bceef225\library\panic_abort\src\lib.rs:32
[...]

PR #275, opened 2023-07-26
Adds new discovered flags to ModuleFlags

ultimaweapon:
Part of #265.


VocalFan:
Wow, 608 needs a lot, haha... Atleast that hopefully means it'll share code with future syscalls.

bayedieng (approved):

PR #279, opened 2023-07-28
Implements tlsinitsize for syscall 608

ultimaweapon:
Part of #265. This also move `VProc` and `VThread` to global scope for supports multi-threading app.

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -10,7 +10,8 @@ members = [
     "param",
     "pfs",
     "pkg",
-    "system"
+    "system",
+    "tls"
 ]
 
 default-members = ["core", "kernel"]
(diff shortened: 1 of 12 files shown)


bayedieng:
Wouldn't integrating our own tls be kindof beyond the scope of our project? Could using the rusttls crate provide the same results?

ultimaweapon:
Our TLS implementation is a [Thread-local storage](https://en.wikipedia.org/wiki/Thread-local_storage), not [Transport Layer Security](https://en.wikipedia.org/wiki/Transport_Layer_Security) 😂.

bayedieng:
Omg lol was so confused reading this PR initially I was wondering why I did not see crypto stuff.

bayedieng:
Also, wanted to ask, how are you determining the validity of the syscall implementation? Is there a document you are referring to get ps4 syscalls if so which one?

bayedieng (approved):

ultimaweapon:
@bayedieng by doing a reverse engineering on the PS4 kernel. The kernel can be dumped from a jailbroken PS4 but some information will be messed up, that why I created #273. Despite some information is messed up it is still very useful.

ultimaweapon:
BTW you can find a dump from someone on the internet, although it is not a recent version but it is still very useful.

Issue #300, opened 2023-08-12
Revises symbol resolver


ultimaweapon:
So its behavior is matched with the PS4.

PR #303, opened 2023-08-14
Moves symbol resolver into a dedicated struct

ultimaweapon:
Part of #300. This also add macOS screenshot.


VocalFan:
> > What resolution do you use or what method to capture the window?
> 
> What do you mean?

Well, all the windows in your Screenshot image are the same size...

ultimaweapon:
I reset the configurations and run it with the default value.

bayedieng (approved):

PR #307, opened 2023-08-20
Implements RTLD list_main

ultimaweapon:
Part of #265 and #300.

### DIFF
--- a/src/gmtx/src/lib.rs
+++ b/src/gmtx/src/lib.rs
@@ -56,14 +56,6 @@ pub struct GroupMutex<T> {
 }
 
 impl<T> GroupMutex<T> {
-    pub const fn new(group: Arc<MutexGroup>, value: T) -> Self {
-        Self {
-            group,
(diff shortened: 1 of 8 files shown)


VocalFan:
So, I only have 1 game I can test Syscall 608 on for... Well:

1. Syscall 599 Reloc 6

The Division
Amplitude
Mighty No. 9
Project Diva X

2. SDK < 0x5000000

Persona 5: Dancing in Starlight
Homebrew Store

3. (S)ELF with no PT_SCE_RELRO

WE ARE DOOMED

Only one that reaches Syscall 608:

Persona 5 Royal

ultimaweapon:
The first one is depend on #300. For the rest I'll take a look when I finished #300 if no one has not taken on it yet.

VocalFan:
Yeah, just saying I won't really have many games to test for 608.

bayedieng (approved):

Issue #302, opened 2023-08-14
Resolve symbol with SDK version < 0x5000000


VocalFan:
A few PS4 applications I have now encounter this error, specifically:

Homebrew Store, and Persona 5: Dancing in Starlight. Both use version `0x04508101`

```
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000002f30:kernel\src/main.rs:83
Starting Obliteration Kernel.
System directory    : G:\obSTORAGE\dev_sys
Game directory      : G:\obSTORAGE\dev_hdd0\CUSA12380
Debug dump directory: C:\Users\Windows Games\AppData\Local\OBHQ\Obliteration\kernel
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000002f30:kernel\src/main.rs:96
Initializing arc4random.
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000002f30:kernel\src/main.rs:100
Initializing LLVM.
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000002f30:kernel\src/main.rs:106
Initializing file system.
++++++++++++++++++ I [00:00:00:00:006]:0x0000000000002f30:kernel\src/main.rs:117
Initializing memory manager.
++++++++++++++++++ I [00:00:00:00:006]:0x0000000000002f30:kernel\src/main.rs:120
Page size is             : 4096
Allocation granularity is: 65536
++++++++++++++++++ I [00:00:00:00:006]:0x0000000000002f30:kernel\src/main.rs:133
Initializing virtual process.
++++++++++++++++++ I [00:00:00:00:007]:0x0000000000002f30:kernel\src/main.rs:138
Initializing runtime linker.
++++++++++++++++++ I [00:00:00:00:012]:0x0000000000002f30:kernel\src/main.rs:149
Application   : /mnt/sandbox/CUSA12380_000/app0/eboot.bin
Image format  : SELF
Image type    : ET_SCE_DYNEXEC
Program 0     : 0x0000000000004000:0x0000000000595a18 => 0x0000000000000000:0x0000000000591a18 PT_LOAD
Program 1     : 0x0000000000598000:0x00000000005d8438 => 0x0000000000594000:0x00000000005d8000 PT_SCE_RELRO
Program 2     : 0x00000000005dc000:0x00000000007dbb58 => 0x00000000005d8000:0x0000000000a41120 PT_LOAD
[...]

PR #308, opened 2023-08-22
Resolves symbol for non-PLT relocations

### LINKED ISSUE #253: Syscall 599: Relocation type 6 is not supported
This error happens on both Amplitude and Project Diva X.

Amplitude log: [amplog.txt](https://github.com/obhq/obliteration/files/12043800/amplog.txt)
Project Diva X: [PDXlog.txt](https://github.com/obhq/obliteration/files/12043801/PDXlog.txt)

ultimaweapon:
Fixes #253.

### DIFF
--- a/src/elf/src/reloc.rs
+++ b/src/elf/src/reloc.rs
@@ -46,6 +46,7 @@ pub struct Relocation {
 impl Relocation {
     pub const R_X86_64_NONE: u32 = 0;
     pub const R_X86_64_64: u32 = 1;
+    pub const R_X86_64_GLOB_DAT: u32 = 6;
     pub const R_X86_64_JUMP_SLOT: u32 = 7;
     pub const R_X86_64_RELATIVE: u32 = 8;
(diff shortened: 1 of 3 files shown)


VocalFan:
For some reason, Linux builds are getting stuck on the Qt install, just had that happen on my Obliteration repository as well! Either way, I'll test this when I get home.

VocalFan:
Yeah, you kinda have to just retry until it works

VocalFan:
From my list earlier:

1. Syscall 599 Reloc 6
The Division
Amplitude
Mighty No. 9
Project Diva X

-----

All games now are a part of issue #302

ultimaweapon:
Seems like we need to move from https://github.com/jurplel/install-qt-action. Have too many problems recently.

VocalFan:
> Seems like we need to move from https://github.com/jurplel/install-qt-action. Have too many problems recently.

Looking into it, it seems like the servers that the Github Linux uses could be bugging out. Other repos are having issues with other packages.

PR #331, opened 2023-09-08
Added Stack Protection

VocalFan:
After some trial and error, I implemented a Stack Guard to help with stack overflows.

(Also used RustFMT and Clippy so if you see misc changes, you know why)

### DIFF
--- a/src/kernel/src/ee/llvm/mod.rs
+++ b/src/kernel/src/ee/llvm/mod.rs
@@ -77,7 +77,7 @@ impl LlvmEngine {
 impl ExecutionEngine for LlvmEngine {
     type RunErr = RunError;
 
-    unsafe fn run(&mut self, _arg: EntryArg, _stack: VPages) -> Result<(), Self::RunErr> {
+    unsafe fn run(&mut self, arg: EntryArg, stack: VPages) -> Result<(), Self::RunErr> {
         todo!()
     }
 }
--- a/src/kernel/src/ee/llvm/mod.rs
+++ b/src/kernel/src/ee/llvm/mod.rs
@@ -77,7 +77,7 @@ impl LlvmEngine {
 impl ExecutionEngine for LlvmEngine {
     type RunErr = RunError;
 
-    unsafe fn run(&mut self, arg: EntryArg, stack: VPages) -> Result<(), Self::RunErr> {
+    unsafe fn run(&mut self, _arg: EntryArg, _stack: VPages) -> Result<(), Self::RunErr> {
         todo!()
     }
 }
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -295,7 +295,7 @@ fn exec<E: ee::ExecutionEngine>(mut ee: E, arg: EntryArg) -> ExitCode {
     {
         use crate::memory::Protections;
         match MemoryManager::current().mprotect(
-            stack.as_mut_ptr(),
+            unsafe{stack.as_mut_ptr().add(stack_size - guard_size)} as *mut _,
             guard_size,
             Protections::empty(),
         ) {
@@ -314,7 +314,7 @@ fn exec<E: ee::ExecutionEngine>(mut ee: E, arg: EntryArg) -> ExitCode {
 
         let locked = unsafe {
             VirtualProtect(
-                stack.as_mut_ptr() as *mut _,
+                stack.as_mut_ptr().add(stack_size - guard_size) as *mut _,
                 guard_size,
                 PAGE_NOACCESS,
                 &mut old_protect as *mut u32,
@@ -327,9 +327,6 @@ fn exec<E: ee::ExecutionEngine>(mut ee: E, arg: EntryArg) -> ExitCode {
             return ExitCode::FAILURE;
         }
     }
-
-    // Skip the guard page in the pointer
-    stack.add(guard_size);
 
     // Start the application.
     info!("Starting application.");
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -294,7 +294,7 @@ fn exec<E: ee::ExecutionEngine>(mut ee: E, arg: EntryArg) -> ExitCode {
     #[cfg(unix)]
     {
         use crate::memory::Protections;
-        match MemoryManager::current().mprotect(stack.as_mut_ptr(), guard_size, PROT_NONE) {
+        match MemoryManager::current().mprotect(stack.as_mut_ptr(), guard_size, Protections::NONE) {
             Ok(_) => (),
             Err(e) => {
                 error!(e, "Guard protection failed");
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -295,7 +295,7 @@ fn exec<E: ee::ExecutionEngine>(mut ee: E, arg: EntryArg) -> ExitCode {
     {
         use crate::memory::Protections;
         match MemoryManager::current().mprotect(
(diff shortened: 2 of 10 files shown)


VocalFan:
Windows is sad in the runner.

VocalFan:
@ultimaweapon Tell me if Linux exploded or not when running a game.

src/kernel/src/memory/page.rs
```diff
@@ -12,6 +12,14 @@ impl<'a> VPages<'a> {
...
     }
 
+    pub fn add(&self, offset: usize) -> Self {
+        Self {
+            mm: self.mm,
+            ptr: unsafe { self.ptr.add(offset) },
+            len: self.len - offset,
+        }
+    }
+
```
ultimaweapon, line 22: This will cause `munmap` multiple times on the same region, which is a bug.

src/kernel/src/memory/mod.rs
```diff
@@ -513,6 +513,7 @@ bitflags! {
     /// Flags to tell what access is possible for the virtual page.
     #[derive(Debug, Clone, Copy, PartialEq)]
     pub struct Protections: u32 {
+        const NONE = 0x00000000;
```
ultimaweapon, line 516: Bit flag with zero value cannot be tested to see if it is present in the set.

src/kernel/src/memory/mod.rs
```diff
@@ -533,6 +534,10 @@ impl Protections {
     fn into_host(self) -> std::ffi::c_int {
         use libc::{PROT_EXEC, PROT_NONE, PROT_READ, PROT_WRITE};
 
+        if self.contains(Self::NONE) {
+            return PROT_NONE; // IF None is assigned, then there will be no other protection flags
+        }
+
```
ultimaweapon, line 540: See my previous comment.

src/kernel/src/ee/llvm/mod.rs
```diff
@@ -77,7 +77,7 @@ impl LlvmEngine {
 impl ExecutionEngine for LlvmEngine {
     type RunErr = RunError;
 
-    unsafe fn run(&mut self, arg: EntryArg, stack: VPages) -> Result<(), Self::RunErr> {
+    unsafe fn run(&mut self, _arg: EntryArg, _stack: VPages) -> Result<(), Self::RunErr> {
```
ultimaweapon, line 80: @bayedieng might be currently working here so rename something here may cause a problem.

ultimaweapon (changes requested):

src/kernel/src/memory/mod.rs
VocalFan: Eh, was just following how FreeBSD uses 0x00 as None, plus I don't see a way to give the Linux page Prot_None without this.

ultimaweapon: You can use `Protections::empty()` to achieve this.

src/kernel/src/memory/page.rs
VocalFan: What would you say would be the best method to fixing this?

src/kernel/src/memory/page.rs
```diff
@@ -12,6 +12,11 @@ impl<'a> VPages<'a> {
         Self { mm, ptr, len }
     }
 
+    pub fn add(&mut self, offset: usize) {
+        self.ptr = unsafe { self.ptr.add(offset) };
+        self.len -= offset;
+    }
+
```
ultimaweapon, line 19: This will cause a memory leak because `munmap` will not apply to the front of allocation.

ultimaweapon (changes requested):

src/kernel/src/memory/page.rs
VocalFan: ...What do I do then? XD

ultimaweapon: You can allocate the stack with the original size (not adding an additional page) and just change the protection on the last page without adjusting the pointer to skip the guard page.

VocalFan: ...Yeah my brain ain't working here.

VocalFan: I tried to see what you were getting at, check last commit XD

VocalFan:
Found out why it's failing. Runner image just got updated.

src/kernel/src/main.rs
```diff
@@ -267,24 +267,67 @@ fn main() -> ExitCode {
...
+        }
+    }
+
+    #[cfg(windows)]
+    {
+        use windows_sys::Win32::System::Memory::{VirtualProtect, PAGE_NOACCESS};
+        let mut old_protect: u32 = 0;
+
+        let locked = unsafe {
+            VirtualProtect(
```
ultimaweapon, line 316: Why you don't use `MemoryManager::mprotect`?

ultimaweapon (changes requested):

src/kernel/src/main.rs
VocalFan: Was just going off of how you used VirtualProtect for Windows in other pieces of code as well.

VocalFan:
Welp, trying to use Mprotect gives me 
```
++++++++++++++++++ E [00:00:00:02:122]:0x000000000000f01c:kernel\src/main.rs:302
Guard protection failed: addr is not aligned.
```

VocalFan:
And yet the Windows method seems to work just fine...

VocalFan:
Oh, no wonder why. We move addr to go past the guard, but that makes the check get angry. Removing the check, it works again.

ultimaweapon:
The error you encountered is correct behavior. You need to pass the lowest address of the stack without adjusting it.

VocalFan:
Bruh

VocalFan:
*Oh, make it so you don't have .add, just make the last part of the page protected!*

*Oh, use mprotect instead of VirtualPage from windows-*

*Oh, make all this happen without changing addr*

VocalFan:
As I hit my head against the wall trying to figure out how to make this work, watch this: https://www.youtube.com/watch?v=c7gp6HeqVfI

ultimaweapon:
Looks like you don't understand the layout of the stack. Better to read this first: https://en.wikipedia.org/wiki/Stack-based_memory_allocation

VocalFan:
Ahaha, I tried again, but it crashes when trying to use asm!

Exception has occurred: Exception
Exception 0xc0000005 encountered at address 0x7ff79f5bf92d: Access violation writing location 0x126ea2e3fd8

VocalFan:
I think I fixed the issue.

src/kernel/src/main.rs
```diff
@@ -267,24 +267,43 @@ fn main() -> ExitCode {
...
         Ok(v) => v,
         Err(e) => {
-            error!(e, "Allocate failed");
+            error!(e, "Stack allocation failed");
             return ExitCode::FAILURE;
         }
     };
 
+    // Calculate the aligned address for the guard page
+    let guard_start = (stack.as_mut_ptr() as usize + guard_size - 1) & !(guard_size - 1);
```
ultimaweapon, line 294: Why guard the page before the last page?

ultimaweapon (changes requested):

src/kernel/src/main.rs
VocalFan: Oh, my b. Seems I accidentally put it in the wrong spot x3

ultimaweapon: All you need to do here is passing `stack.as_mut_ptr()` directly as a first argument of `mprotect`.

ultimaweapon (approved):

PR #336, opened 2023-09-12
Add Default case for regmgr_call.

### LINKED ISSUE #332: Syscall 532, Implement Default
Default would allow 0x27 and any types past 0x36 to work.

VocalFan:
Added Default case for regmgr_call, allowing types 0x27 and any type past 0x36 to work. I implemented it in a way that it won't Default any values that DO have a special case (0x0-> 0x26 and 0x28 -> 0x36) are special cases, so I used (0x27 | 0x37..=u32::max)

Fixes #332


src/kernel/src/syscalls/mod.rs
```diff
@@ -243,6 +243,9 @@ impl Syscalls {
                     todo!("regmgr_call with matched key = {key:#x}");
                 }
             }
+            // Pass default value for any values that aren't given a unique case
+            0x27 | 0x37..=u32::MAX => -2146631143,
+            // Any missing cases TODO
```
ultimaweapon, line 248:
```c
    case 0x37:
        if (DAT_ffffffffa2830830 == 0) {
LAB_ffffffffa08df0f3:
            key = 0x800d0201;
        }
        else {
            if ((DAT_ffffffffa2830834 == 0) || (DAT_ffffffffa2830800 == 1)) {
                key = FUN_ffffffffa08ced20();
                goto LAB_ffffffffa08df4bf;
            }
LAB_ffffffffa08df0dd:
            key = 0x800d020d;
            if ((int)DAT_ffffffffa2830834 < 0) {
                key = DAT_ffffffffa2830834;
            }
        }
LAB_ffffffffa08df4cc:
        rVar8 = 0;
LAB_ffffffffa08df4d6:
        FUN_ffffffffa08eba50(5,rVar8,key & 0xffff);
        break;
```

ultimaweapon (changes requested):

src/kernel/src/syscalls/mod.rs
VocalFan: Damn, I was one off XD

VocalFan:
Not sure where you got 0x37 though, as that wasn't in sys_regmgr_call

VocalFan:
Seeing as:

```Cpp
        case 0x36uLL:
          v33 = sceRegMgrPrivateStorageIsNvsChange();
          break;
        default:
          v33 = -2146631143;
          break;
      }
```

ultimaweapon:
If you take those from the decompiled it is an old one.

src/kernel/src/syscalls/mod.rs
```diff
@@ -243,6 +243,9 @@ impl Syscalls {
                     todo!("regmgr_call with matched key = {key:#x}");
                 }
             }
+            // Pass default value for any values that aren't given a unique case
+            0x27 | 0x38..=u32::MAX => -2146631143,
+            // Any missing cases TODO
```
ultimaweapon, line 248:
```c
    case 0x38:
        if (DAT_ffffffffa2830830 == 0) goto LAB_ffffffffa08df0f3;
        if ((DAT_ffffffffa2830834 != 0) && (DAT_ffffffffa2830800 != 1)) goto LAB_ffffffffa08df0dd;
        key = FUN_ffffffffa08d04d0();
        goto LAB_ffffffffa08df4bf;
```

ultimaweapon (changes requested):

src/kernel/src/syscalls/mod.rs
VocalFan: Stop going one by one because the FW leak only goes to 36 XD

ultimaweapon: If you mean the one that you send it to me it is not a source leak but an old binary leak that was decompiled by IDA Pro. You should follow the most recent one that you have, which is 9.0; and use the leaked one as a supplement.

VocalFan: Ah, I see 0x39 is the highest so far.

VocalFan:
Fixed it by looking at and searching in 9.00's regmgr_call. Highest value was 0x39

VocalFan:
Well, that confirms the hex is indeed -2146631143

ultimaweapon (approved):

PR #338, opened 2023-09-13
Fix wrong parameter in error macro for non-x64

bayedieng:
We should really look into getting github actions for m1 macs these type of issues are hard to catch due to conditional compilation.


ultimaweapon:
Seems like the only possible solution right now is cross-compiling because GitHub does not provide a Mac with Apple Silicon yet.

bayedieng:
Yeah, seems like the best solution for now i've tried arm mac to intel mac cross-compilation before  but not the inverse, should work regardless.

ultimaweapon (approved):

PR #346, opened 2023-09-14
Improvements to Game List

VocalFan:
As it sounds, added the ability for the user to easily open the system folder!


VocalFan:
Decided to add some changes: The game list now auto scales for the window being resized instead of staying tucked in one spot or overflowing into a scroll bar. I also decided to fix the alphabetical order for the context menu so Settings was under Open Game Folder

VocalFan:
And now added name sorting for games.

src/game_models.cpp
```diff
@@ -73,3 +75,10 @@ QVariant GameListModel::data(const QModelIndex &index, int role) const
         return QVariant();
     }
 }
+
+void GameListModel::sortNames()
+{
+    std::sort(m_items.begin(), m_items.end(), [](Game *a, Game *b) {
+        return a->name() < b->name();
+    });
+}
```
ultimaweapon, line 84: Have to read this https://doc.qt.io/qt-6/model-view-programming.html?

ultimaweapon (changes requested):

src/game_models.cpp
VocalFan: Nope, but I just poked until I got something working lol

ultimaweapon: Then better to remove this because GameListModel is more complicated than what it looks.

ultimaweapon (approved):

src/game_models.cpp
VocalFan: Looking back at the URL, I don't quite see the concern. All it does is reorder the entries of games that way they go in alphabetical order. The game's folder path and all still matches the game.

So, if I had:

Game1
CUSA00001

Ame2
CUSA00002

It would sort as:

Ame2
CUSA00002

Game1
CUSA00001

You won't get an issue such as:

Ame2
CUSA00001 (Game1's folder)

Game1
CUSA00002 (Ame2's folder)

ultimaweapon: How does the widget know it needs to update because the order of the items was changed?

VocalFan: Hmm? You don't have to explicitly tell the widget anything. It reflects off of the model. If the model gets sorted, the widget's options get sorted as well.

ultimaweapon: No it need some mechanism to tell the widget that the model has been updated. It you think it is happen automagically that mean you don't understand the how it is working yet.

VocalFan: > No it need some mechanism to tell the widget that the model has been updated. It you think it is happen automagically that mean you don't understand the how it is working yet.

Despite it working already? Ugh, fine... I'll add resets...

ultimaweapon: Because it is C++. Even if it work does not mean it is working correctly.

PR #347, opened 2023-09-16
Add name sorting.

VocalFan:
Fixed an issue to make sure the Widget refreshes after sorting, also fixed an issue with icons not being uniformly sized. Here's an example of the sorting:

Before:
image

After:
image


ultimaweapon (changes requested):
This is not a correct way to do it. Better to understand how Model/View in Qt is working before implement this.

ultimaweapon:
Wait, why this contains the commits that already merged?

VocalFan:
> Wait, why this contains the commits that already merged?

Probs because it's the same branch just... Updated?

ultimaweapon:
Should not be a problem if GitHub does not show any conflicts here.

VocalFan:
I think we should uh, fix that issue if an app has an abnormally large icon.

ultimaweapon:
What do you mean by a large icon?

VocalFan:
Look at the before and after. Since Itemzflow has a 1000px logo, it was way larger than the other logos that were 512px. Since the PS4 ui just downscales the icon to match the normal size, I did so as well.

VocalFan:
Don't merge btw, gotta fix an issue that capital letters are prioritized. In unicode and ascii, capital letters come before lower letters in order.

ultimaweapon (approved):

Issue #369, opened 2023-09-28
Unimplemented Syscall 5


VocalFan:
Syscall 5, also known as [open](https://man.freebsd.org/cgi/man.cgi?query=open&apropos=0&sektion=2&manpath=FreeBSD+9.1-RELEASE&arch=default&format=html)

The source code is here: https://cgit.freebsd.org/src/tree/sys/kern/vfs_syscalls.c?h=releng/9.1#n1038

PR #373, opened 2023-09-29
Implements syscall 56

### LINKED ISSUE #368: Unimplemented Syscall 56
Syscall 56, also known as [revoke](https://man.freebsd.org/cgi/man.cgi?query=revoke&apropos=0&sektion=2&manpath=FreeBSD+9.1-RELEASE&arch=default&format=html)

The source code is here: https://cgit.freebsd.org/src/tree/sys/kern/vfs_syscalls.c?h=releng/9.1#n4274

ultimaweapon:
Fixes #368.


VocalFan:
Games now reach #369

VocalFan:
I'll be honest Ultima, I'm still surprised at how my fest of returning all those syscalls as 0s let me get that far lmao. *Of course, it isn't proper behavior, but it let us see what future syscalls we had to implement.

ultimaweapon:
All of those might not be required in the beginning. For example, if you change the output of `is_in_sandbox` to 1 it will go to a different syscall.

VocalFan:
> All of those might not be required in the beginning. For example, if you change the output of `is_in_sandbox` to 1 it will go to a different syscall.

I might do that and make issues for all of them in case anybody wants to implement the syscalls.

VocalFan:
Also, glad you're deciding to take a more verbose approach to logging now. Seeing as Syscalls kinda said nothing a week ago.

VocalFan:
Also, why are you defining your own "no such file or directory" error when we have an error code for that already? (Enoent)

bayedieng (approved):

ultimaweapon:
The errno is for returning to the PS4.

PR #412, opened 2023-10-23
Add other signals

VocalFan:
This is meant to be for when syscall 416 is called and the Signals are listed.

These Signals were taken from ORBISPlugin_2015


ultimaweapon:
What is ORBISPlugin_2015?

VocalFan:
> What is ORBISPlugin_2015?

It's from the SDK-Tools.

ultimaweapon (approved):

Issue #401, opened 2023-10-17
Detroit: Become Human


VocalFan:
```
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000004808:kernel\src/main.rs:92
Starting Obliteration Kernel.
System directory    : D:\obSTORAGE\dev_sys
Game directory      : D:\obSTORAGE\dev_hdd0\CUSA08344
Debug dump directory: C:\Users\ghost\AppData\Local\OBHQ\Obliteration\kernel
++++++++++++++++++ I [00:00:00:00:001]:0x0000000000004808:kernel\src/main.rs:125
Page size             : 4096
Allocation granularity: 65536
Main stack            : 0x22f545b4000:0x22f547b4000
++++++++++++++++++ I [00:00:00:00:002]:0x0000000000004808:kernel\src/main.rs:188
Initializing runtime linker.
++++++++++++++++++ E [00:00:00:01:777]:0x0000000000004808:kernel\src/main.rs:193
Initialize failed: cannot setup /mnt/sandbox/CUSA08344_000/app0/eboot.bin -> cannot build a trampoline for 0x15c7627 -> the remaining workspace is not enough.
```

I've tested on the pre-165 syscall build of Obliteration, for which the game reaches `Unimplemented syscall 165`

VocalFan:
@ultimaweapon Also occurs on two other games...

Tom Clancy The Division:
```
++++++++++++++++++ I [00:00:00:00:003]:0x0000000000007c14:kernel\src\discord_presence\mod.rs:9
Initializing Discord rich presence.
++++++++++++++++++ I [00:00:00:00:163]:0x0000000000007c14:kernel\src/main.rs:99
Starting Obliteration Kernel.
System directory    : G:\obSTORAGE\dev_sys
Game directory      : G:\obSTORAGE\dev_hdd0\CUSA01810
Debug dump directory: C:\Users\Windows Games\AppData\Local\OBHQ\Obliteration\kernel
++++++++++++++++++ I [00:00:00:00:163]:0x0000000000007c14:kernel\src/main.rs:132
Page size             : 0x1000
Allocation granularity: 0x10000
Main stack            : 0x21eb09b4000:0x21eb0bb4000
++++++++++++++++++ I [00:00:00:00:164]:0x0000000000007c14:kernel\src/main.rs:199
Initializing runtime linker.
++++++++++++++++++ E [00:00:00:00:570]:0x0000000000007c14:kernel\src/main.rs:204
[...]

PR #417, opened 2023-10-26
Refactored errors in kernel main, fixed grammar

SuchAFuriousDeath:
The match statements in the main function looked kinda awkward, so I decided to do a little refactor

### DIFF
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -48,7 +48,7 @@ fn main() -> Result<(), KernelError> {
 
     // Load arguments.
     let args = if std::env::args().any(|a| a == "--debug") {
-        let file =  File::open(".kernel-debug")?;
+        let file = File::open(".kernel-debug")?;
 
         serde_yaml::from_reader(file)?
     } else {
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -22,7 +22,6 @@ use std::fmt::Debug;
 use std::fs::{create_dir_all, remove_dir_all, File};
 use std::io::Write;
 use std::path::PathBuf;
-use std::process::{Termination, ExitCode};
 use std::sync::Arc;
 use thiserror::Error;
 
@@ -283,18 +282,16 @@ enum KernelError {
 
 impl Debug for KernelError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        use std::error::Error;
-
-        write!(f, "{self}");
+        write!(f, "{self}")?;
 
         let mut i = self.source();
 
         while let Some(v) = i {
-            write!(f, " -> {}", v).unwrap();
+            write!(f, " -> {}", v)?;
             i = v.source();
         }
 
-        write!(f, "")
+        Ok(())
     }
 }
 
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -260,37 +260,37 @@ fn join_thread(thr: Thread) -> Result<(), std::io::Error> {
 
 #[derive(Error, Termination)]
 enum KernelError {
-    #[error("Failed to open .kernel-debug: {0}")]
+    #[error("Failed to open .kernel-debug -> {0}")]
     FailedToOpenKernelDebug(#[from] std::io::Error),
 
-    #[error("Failed to parse .kernel-debug: {0}")]
+    #[error("Failed to parse .kernel-debug -> {0}")]
     FailedToParseKernelDebug(#[from] serde_yaml::Error),
 
-    #[error("Cannot get CPU time limit: {0}")]
+    #[error("Cannot get CPU time limit -> {0}")]
     VirtualProcessInitialzationFailed(#[from] crate::process::VProcError),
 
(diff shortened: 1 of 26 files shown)


src/kernel/src/main.rs
```diff
@@ -40,26 +43,14 @@ mod syscalls;
 mod sysctl;
 mod ucred;
 
-fn main() -> ExitCode {
+fn main() -> Result<(), KernelError> {
```
ultimaweapon, line 46: Could you show some examples of the output? Not sure if this supports nested error.

SuchAFuriousDeath: Error: Failed to open .kernel-debug: No such file or directory (os error 2)

SuchAFuriousDeath: It seems that it can nest arbitrarily. VocalFan can confirm

SuchAFuriousDeath: By the way, all those match statements all over the codebase could be refactored with #[from] instead of #[source]

ultimaweapon: > Error: Failed to open .kernel-debug: No such file or directory (os error 2)

This one can't be tested to see if it is support nested error. You need something like this https://github.com/obhq/obliteration/issues/401#issue-1947775969.

VocalFan: I tested it in Discord, we made it able to match issue 401's error, which is what the later Commits are, @ultimaweapon

SuchAFuriousDeath: Yup. VocalFan tested that exact one and it worked

SuchAFuriousDeath: So if I understand correctly, the issue is that with external errors we have to call ->source() to the nested one and the error! macro did that? If so, I could implement the Termination trait manually and use error!

ultimaweapon: Correct. With `Error::source` it is guaranteed that we will be able to get the inner error from any errors.

SuchAFuriousDeath: Should I also use #[from] to simplify all the other errors outside of main? @ultimaweapon. I think it would be more readable

VocalFan:
image

VocalFan:
Besides from FMT, seems to work fine!

src/kernel/src/main.rs
ultimaweapon: How we will handle this case?

```rust
fn foo() -> Result<(), FooError> {
    let file = File::open("abc")?;
    file.read(&mut buf)?;
}

#[derive(Debug, Error)]
enum FooError {
    Bar(#[from] std::io::Error)
}
```

SuchAFuriousDeath: What's the issue here?

SuchAFuriousDeath: I already looked at all the existing code and there are definitely some issues, but I don't quite see one here. One of the issues is that more that one variant contains the same error and another one is that a lot of variants contain more info than just the error. This could be fixed by .map_err, but I'm not quite sure if that's your style of code, that's up to you. There's also a lot of opportunities to use .ok_or on Option but again, not sure if you consider that better than a match statement. You tell me

ultimaweapon: The problem with `#[from]` here is easy to accidentally return the wrong variant like my above example that a single variant will represent both error for `open` and `read`.

I'm fine with both `map_err` and `ok_or` as long as the call chain is not too long and it does not require cloning a heap allocated value.

SuchAFuriousDeath: I see. I don't suppose we can do anything about that, other than just watch out.

ultimaweapon:
Sorry for the late review. Just came back from my vacation. I hope I can take a look tonight.

SuchAFuriousDeath:
I still need to fix the checks here

ultimaweapon:
Not sure if you are waiting for the review. What I understand is you will fix the checks first.

SuchAFuriousDeath:
> Not sure if you are waiting for the review. What I understand is you will fix the checks first.

I didn't really have time to work on this because of work, I'll check during the weekend

ultimaweapon:
Don't worry. Just asking to see if you are waiting for me so I can do an appropriate action.

SuchAFuriousDeath:
By the way, I would be delighted if you were willing to share some RE information so that people trying to join the project don't have to reinvent the wheel. Also, it might be useful to put this link to the Github discussion with the syscall information in the README (https://github.com/obhq/obliteration/discussions/292 ) as it's pretty crutial information and it's easy to miss.

ultimaweapon:
The problem is sharing that information is hard because it is tool specific data it was not designed for collaboration.

ultimaweapon:
The problem is we can't publish Ghidra data because:

- It is copyrighted data.
- It is not Git friendly.

ultimaweapon:
The best thing we can do is something like this: https://github.com/obhq/reverse-engineering

Create an issue there if you need any information like function name at the specific offset.

ultimaweapon:
@SuchAFuriousDeath are you still working on this? If not better to close this PR.

SuchAFuriousDeath:
I think I'll close this and try from scratch. I tried to do too many things at once anyway

PR #422, opened 2023-11-03
xsave64 alignment fix

### LINKED ISSUE #419: SIGSEGV on xsave64 instruction
image
This instruction causes a SIGSEGV for me. It's located in the patched ELF file. I think my xsave_area is too big (2696 bytes). When my friend with the same OS sends me his compiled kernel binary, the error happens for me when I run it, but not for him, his xsize_area is 1088. When I put in 1088 for myself, it passes the xsave64 instruction and crashes on one of the next instructions.

The assembly instruction corresponds to this line - https://github.com/obhq/obliteration/blob/5bdc59ec6b426963fa7222ac6280af47b893c0ab/src/kernel/src/ee/native/mod.rs#L526-L533

Maybe the xsave area is too big to subtract from rsp? I'm not sure, I hope someone has in idea of what's happening here

SuchAFuriousDeath:
Fixes #419 for me and the other commenter on the issue who suggested the fix.

And also, the Intel developer manual does actually state that the address should be 64-byte aligned.

### DIFF
--- a/src/kernel/src/ee/native/mod.rs
+++ b/src/kernel/src/ee/native/mod.rs
@@ -521,7 +521,7 @@ impl NativeEngine {
         asm.add(rbx, 14 * 8).unwrap(); // Output placeholder.
 
         // Save x87, SSE and AVX states.
-        let xsave = (self.xsave_area + 15) & !15;
+        let xsave = (self.xsave_area + 63) & !63;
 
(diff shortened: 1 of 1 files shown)


ultimaweapon:
`jurplel/install-qt-action` hit us again. We should really consider move away from this because we have too many problems with it.

ultimaweapon (approved):

PR #423, opened 2023-11-03
Add relocation constants

SuchAFuriousDeath:
Deduced from strings in the binary

### DIFF
--- a/src/elf/src/reloc.rs
+++ b/src/elf/src/reloc.rs
@@ -46,10 +46,16 @@ pub struct Relocation {
 impl Relocation {
     pub const R_X86_64_NONE: u32 = 0;
     pub const R_X86_64_64: u32 = 1;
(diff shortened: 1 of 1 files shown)


ultimaweapon:
Where did you get these from?

SuchAFuriousDeath:
If you search for strings in ghidra and filter them with 'R_86', and then get where the resulting string are used (most of them), it's a function with a switch statement where the cases clearly correspond to relocation constants, some of which we are missing.

ultimaweapon:
BTW do you want to join the organization?

SuchAFuriousDeath:
You mean Obliteration? I'd love to.

ultimaweapon:
Nope. I mean:

OrganizationXIII

BTW invitation has been sent. You can check organization page to see the invitation if you don't see in the notification.

ultimaweapon (approved):

PR #424, opened 2023-11-04
Removes jurplel/install-qt-action

ultimaweapon:
We got too many problems with it so better to remove it.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -5,57 +5,72 @@ jobs:
   build:
     name: Linux
     runs-on: ubuntu-22.04
-    env:
-      CMAKE_BUILD_PARALLEL_LEVEL: '2'
     steps:
     - name: Checkout source
       uses: actions/checkout@v4
+    - name: Install System Packages
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y flatpak flatpak-builder
+    - name: Add Flathub
+      run: flatpak remote-add --user flathub https://flathub.org/repo/flathub.flatpakrepo
     - name: Generate cache keys
       run: |
-        echo "cargo=${{ runner.os }}-cargo" >> $GITHUB_OUTPUT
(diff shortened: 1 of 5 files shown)


VocalFan:
Thanks bestieeeeeeeeee

VocalFan:
Also woah, flatpak builds!?

ultimaweapon:
Because I can reuse the previous configurations without writing from scratch.

VocalFan:
...This overall is fixing a LOT of pain we had with Actions... Removing god awful LLVM times for Windows, adding Flatpak builds, and getting rid of that broken Qt action.

ultimaweapon:
Looks like LLVM installer for Windows does not provide `llvm-config.exe`.

VocalFan:
@ultimaweapon Can't we cache the entirety of the LLVM build to speed up the process? After all, LLVM really only has to rebuild when there's a new version or conditions change. Which if they do, the build tool automatically replaces the old version.

Y'know... The lib/ folder?

VocalFan:
image

ultimaweapon:
Strange. Why the cache is not found.

VocalFan:
81 minutes, rerunning to see cache time diff.

VocalFan:
And down to 1 minute! But I think that can be even quicker... Maybe. @ultimaweapon Do you still need to do the Download LLVM step on "LLVM for Windows" even though  you already have the compiled lib? If not, that'd make it basically be instant.

ultimaweapon:
That one is required to generate a key.

VocalFan:
Ah, I see that now. Either way Ultima, I'd say that was a damn good suggestion haha. And over all, good work! Got any other changes?

VocalFan (approved):

ultimaweapon:
Just got another idea. Instead of hashing the downloaded file we can hash the URL instead. But that will take another hour so let's use the current version for now and implement this together with LLVM upgrade.

PR #429, opened 2023-11-06
Add types from Orbis.

VocalFan:
Identified types so far: Program, Relocation, and Symbol.

### DIFF
--- a/src/kernel/src/rtld/resolver.rs
+++ b/src/kernel/src/rtld/resolver.rs
@@ -271,6 +271,7 @@ impl<'a, E: ExecutionEngine> SymbolResolver<'a, E> {
                     || ((ty == Symbol::STT_NOTYPE
                         || ty == Symbol::STT_OBJECT
                         || ty == Symbol::STT_FUNC
+                        || ty == Symbol::STT_COMMON
                         || ty == Symbol::STT_ENTRY)
                         && sym.value() != 0))
                     && (sym.shndx() != 0
@@ -383,7 +384,7 @@ impl<'a, E: ExecutionEngine> SymbolResolver<'a, E> {
         let ty = sym.ty();
 
         match ty {
-            Symbol::STT_NOTYPE | Symbol::STT_OBJECT | Symbol::STT_FUNC | Symbol::STT_ENTRY => {
+            Symbol::STT_NOTYPE | Symbol::STT_OBJECT | Symbol::STT_FUNC | Symbol::STT_COMMON | Symbol::STT_ENTRY => {
                 if sym.value() == 0 {
                     return false;
                 }
--- a/src/kernel/src/rtld/resolver.rs
+++ b/src/kernel/src/rtld/resolver.rs
@@ -271,7 +271,6 @@ impl<'a, E: ExecutionEngine> SymbolResolver<'a, E> {
                     || ((ty == Symbol::STT_NOTYPE
                         || ty == Symbol::STT_OBJECT
                         || ty == Symbol::STT_FUNC
-                        || ty == Symbol::STT_COMMON
                         || ty == Symbol::STT_ENTRY)
                         && sym.value() != 0))
                     && (sym.shndx() != 0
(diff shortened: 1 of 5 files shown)


src/kernel/src/rtld/resolver.rs
```diff
@@ -271,6 +271,7 @@ impl<'a, E: ExecutionEngine> SymbolResolver<'a, E> {
                     || ((ty == Symbol::STT_NOTYPE
                         || ty == Symbol::STT_OBJECT
                         || ty == Symbol::STT_FUNC
+                        || ty == Symbol::STT_COMMON
```
ultimaweapon, line 274: Where is this check on the PS4?

src/kernel/src/rtld/resolver.rs
```diff
@@ -383,7 +384,11 @@ impl<'a, E: ExecutionEngine> SymbolResolver<'a, E> {
         let ty = sym.ty();
 
         match ty {
-            Symbol::STT_NOTYPE | Symbol::STT_OBJECT | Symbol::STT_FUNC | Symbol::STT_ENTRY => {
+            Symbol::STT_NOTYPE
+            | Symbol::STT_OBJECT
+            | Symbol::STT_FUNC
+            | Symbol::STT_COMMON
```
ultimaweapon, line 390: Same here.

src/kernel/src/rtld/resolver.rs
VocalFan: It's actually in the SDK, I'll get it when home if ya want.

VocalFan:
```
const char *__fastcall sub_14003B270(__int64 a1, int a2)
{
  const char *result; // rax
  if ( (unsigned int)(a2 - 13) <= 2 )
    return "Unknown processor specific";
  switch ( 0x40000000 )
  {
    case 0:
      result = "STT_NOTYPE";
      break;
    case 1:
      result = "Object";
      break;
    case 2:
      result = "Function";
      break;
    case 3:
      result = "Section";
      break;
    case 4:
      result = "Source File";
      break;
    case 5:
      result = "STT_COMMON";
      break;
    case 6:
      result = "STT_TLS";
      break;
    case 10:
      result = "STT_LOOS";
      break;
    case 11:
      result = "Module Entry";
      break;
    case 12:
      result = "STT_HIOS";
      break;
    default:
      result = "Unknown type";
      break;
  }
  return result;
}
```

VocalFan: Oh wait, if you mean how its handled, read its description when I defined it.

VocalFan:
```
    /// Symbol labels an uninitialized common block (Treated same as STT_OBJECT)
    pub const STT_COMMON: u8 = 5;
```

ultimaweapon: I mean in the PS4 kernel.

VocalFan: > I mean in the PS4 kernel.

Heck if I know! I'm never up to date with your reverse engineering x3... I'm just going off the Orbis SDK.

ultimaweapon: Then you need to remove this code because this check is not in the kernel.

VocalFan: > Then you need to remove this code because this check is not in the kernel.

Except in all documentation I could find, it's meant to be treated exactly like STT_OBJECT

ultimaweapon: What we do is emulate the PS4 kernel so why should we follow those documents if it is conflict with the kernel?

VocalFan: ...I would doubt this as FreeBSD 9.1 does this. Along with STT_COMMON just being a `a common data object` while STT_OBJECT is a `a data object`. orbis-as.c and orbis-bin.c both contain it. All documentation says that common is treated like object...

ultimaweapon: The only cases that I'll allow this is you show a proof from the PS4 kernel.

VocalFan: Changed -w-

ultimaweapon (approved):

Issue #428, opened 2023-11-05
Unimplemented syscall 594


SuchAFuriousDeath:
According to the PS4 dev wiki, the signature is:
int sys_dynlib_load_prx(const char *moduleFileName, size_t args, const void *argp, uint32_t flags, const SceKernelLoadModuleOpt *pOpt, int *pRes) 

This can be reached by stubbing the not-yet-implemented syscalls and returning ENOENT from Sysctl::exec

red-prig:
This is actually dynamic loading of libraries, because at startup, in addition to the program itself, only libkernel and libcinternal are loaded, everything else is loaded using this system call

ultimaweapon:
Thanks.

SuchAFuriousDeath:
@ultimaweapon I think you'll have to do this one, it looks too complicated for me. Maybe I can do the sysctl issue.

Issue #416, opened 2023-10-23
Not Yet Implemented: sysctl [1, 14, 41]


VocalFan:
```
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 557
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\process\mod.rs:441
Named object 'SceLibcI_FileStderr' (ID = 15) was created with data = 0x2000004c0 and flags = 0x101.
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 202
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 202
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 487
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\process\mod.rs:451
Returning bypass as 0.
++++++++++++++++++ I [00:00:00:01:584]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 340
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 340
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 596
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\process\mod.rs:451
Returning bypass as 0.
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 598
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 605
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\process\mod.rs:451
Returning bypass as 0.
++++++++++++++++++ I [00:00:00:01:585]:0x0000000000001bc0:kernel\src\syscalls\mod.rs:51
Syscall called: 202
[...]

PR #438, opened 2023-11-12
Several Advancements

SuchAFuriousDeath:
Partial implementation of #428 (to the point where it allows us to go further), stubbed #416 and the newly reachable syscalls.
I also added a table of sysctls for a better overview and easier searching in the future.

UPDATE: also added a near-complete implementation of syscall 649


SuchAFuriousDeath:
By the way, I think we could really use a get_module_by_id function on RuntimeLinker (instead of doing self.list.read().iter().find(|m| m.id() == handle) 4 times), but I'm afraid it requires lifetime generics which make my brain hurt.

ultimaweapon (changes requested):
I'm not a fan of subbing because it is likely to cause problems in the future.

SuchAFuriousDeath:
I tend to agree, you can never be sure what's caused by what thing. I should have made this a draft, probably.

SuchAFuriousDeath:
@ultimaweapon Also, what do I have to do to make this PR mergeable?

ultimaweapon:
Better to keep only codes that have fully implementation. For partial implementation it should have a full implementation at least one branch and put `todo` macro on the other branches.

SuchAFuriousDeath:
@ultimaweapon How will we implement /dev/console and stdout btw? Will we combine them into one stream?

ultimaweapon:
My idea is we will combine them to stderr and add another tab on the GUI to show this.

SuchAFuriousDeath:
What about when running just the kernel?

ultimaweapon:
Same because right now the kernel print all its log to stdout. I'm keeping stderr specific for this.

SuchAFuriousDeath:
I got too excited and put way to much in this PR once again.

SuchAFuriousDeath:
The funny thing is, by implementing the devices and write instead of returning letting them fail, we actually hit a wall much earlier :(.

ultimaweapon:
> I got too excited and put way to much in this PR once again.

Yeah you put too much things in a single PR. Better to close this and split it into multiple PR will be better.

PR #444, opened 2023-11-14
Add sysinfo and update LLVM

VocalFan:
Due to us seeming to find OS specific and Hardware specific issues more and more, I decided to add sysinfo to display the Operating System, CPU, and Available/Total RAM. I also updated LLVM from 17.0.2 to 17.0.5 while also adding it where we hash by URL instead of file like Ultima suggested.
```
Operating System    : Windows 11 Pro 22621
CPU Information     : AMD Ryzen 9 3900X 12-Core Processor            
Memory Available    : 53100/65461 MB
```

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -12,11 +12,11 @@ jobs:
     steps:
     - name: Download LLVM
       run: |
-        Invoke-WebRequest -Uri "https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.2/llvm-project-17.0.2.src.tar.xz" -OutFile llvm.tar.xz
-        $hash = (Get-FileHash llvm.tar.xz).Hash
+        $llvmurl = "https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.5/llvm-project-17.0.5.src.tar.xz"
+        Invoke-WebRequest -Uri $llvmurl -OutFile llvm.tar.xz
         unxz llvm.tar.xz
         tar -x --strip-components=1 -f llvm.tar
-        echo "hash=$hash" >> $env:GITHUB_OUTPUT
+        echo "hash=$llvmurl" >> $env:GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


ultimaweapon (changes requested):
You need to separated cache key creation from download step and skip the download step if hash of the URL is already cached.

VocalFan:
Yeah I noticed that too, but a migraine made me go to sleep while I waited for the llvm url to build xD

VocalFan:
Yeah, got it working lol

VocalFan:
MacOS moment

.github/workflows/ci-windows.yml
```diff
@@ -10,16 +10,13 @@ jobs:
...
-        $hash = (Get-FileHash llvm.tar.xz).Hash
-        unxz llvm.tar.xz
-        tar -x --strip-components=1 -f llvm.tar
-        echo "hash=$hash" >> $env:GITHUB_OUTPUT
+        $url = "https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.5/llvm-project-17.0.5.src.tar.xz"
+        echo "llvmurl=$url" >> $env:GITHUB_OUTPUT
       id: download
     - name: Generate cache keys
-      run: echo "dist=${{ runner.os }}-llvm-${{ steps.download.outputs.hash }}" >> $env:GITHUB_OUTPUT
+      run: echo "dist=${{ runner.os }}-llvm-${{ steps.download.outputs.llvmurl }}" >> $env:GITHUB_OUTPUT
```
ultimaweapon, line 19: Better to hash the URL instead of using the URL itself so the key will not be too long.

ultimaweapon (changes requested):

.github/workflows/ci-windows.yml
VocalFan: Not very pretty doing a hash through powershell xS

ultimaweapon (approved):

PR #446, opened 2023-11-15
Implement write, close and dipsw ioctl

SuchAFuriousDeath:


SuchAFuriousDeath:
@ultimaweapon How do we implement /dev/dipsw?

ultimaweapon:
Better to follow how the PS4 implements it.

SuchAFuriousDeath:
Do we need that right now or do we just return Ok(()) the same way it's done for console ioctl?

ultimaweapon:
Try full implementation first if possible. If not possible try at least one branch. If that is still not possible then use the best effort to guess what the PS4 is doing.

SuchAFuriousDeath:
You have any idea what it is and what it does though? From a bit of googling, it seems to be some sort of a dip switch. I doubt it means 'diploma in social work' :D.

ultimaweapon:
Nope.

SuchAFuriousDeath:
This one is very close to being finished. The command we're getting now again checks some 2 functions before writing something (it's referencing some static address), otherwise writing 0, much like kern_proc_sanitizer.

SuchAFuriousDeath:
I don't think our argument to ioctl can be a slice. The FreeBSD code seems be a bit too chaotic for that

SuchAFuriousDeath:
Also, I think the original code was wrong and didn't correspond to the control flow.

src/kernel/src/fs/mod.rs
```diff
@@ -12,17 +12,22 @@ use param::Param;
...
 
+mod dev;
 mod file;
 mod item;
 mod path;
 mod vnode;
 
+//Represents a file descriptor
+pub type Fd = i32;
+
```
ultimaweapon, line 30: Better to make it as a tuple struct instead for strong typing.

src/kernel/src/fs/dev/dipsw.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Error, Debug)]
+enum IoctlErr {
+    Invalid,
+}
+
+impl Display for IoctlErr {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            IoctlErr::Invalid => write!(f, "Invalid ioctl"),
```
ultimaweapon, line 80: Why not using `error` attribute from `thiserror`?

ultimaweapon (changes requested):

src/kernel/src/fs/dev/dipsw.rs
SuchAFuriousDeath: Will do

src/kernel/src/fs/mod.rs
SuchAFuriousDeath: Sorry about that, I though I already did that, I must have used git stash on it or something.

src/kernel/src/fs/dev/dipsw.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl Display for Dipsw {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        Self::PATH.fmt(f)
+    }
+}
+
+#[derive(Error, Debug)]
+enum IoctlErr {
+    #[error("Invalid command")]
```
ultimaweapon, line 73: Here is a guide for error message: https://doc.rust-lang.org/std/error/trait.Error.html

src/kernel/src/syscalls/output.rs
```diff
@@ -44,3 +46,9 @@ impl From<NonZeroI32> for SysOut {
         }
     }
 }
+
+impl From<Fd> for SysOut {
+    fn from(value: Fd) -> Self {
+        value.get().into()
+    }
+}
```
ultimaweapon, line 54: Better to implement this at the same place as `Fd` so this file will not get polluted with every types in our project. This file should contains only implementations for the external types.

src/kernel/src/process/file.rs
```diff
@@ -68,4 +72,22 @@ impl FileDesc {
 
         files.get(fd)?.clone()
     }
+
+    pub fn free(&self, fd: Fd) -> Result<(), SysErr> {
+        if fd.get() < 0 {
+            return Err(SysErr::Raw(EBADF));
+        }
```
ultimaweapon, line 79: This check should be a responsibility of `Fd` type like `str` in Rust that it will never allow non-UTF8.

src/kernel/src/process/file.rs
```diff
@@ -49,17 +53,17 @@ impl FileDesc {
...
         // This should never happened.
         panic!("Too many files has been opened.");
     }
 
     /// See `fget` on the PS4 for a reference.
-    pub fn get(&self, fd: i32) -> Option<Arc<VFile>> {
+    pub fn get(&self, fd: Fd) -> Option<Arc<VFile>> {
         // TODO: Check what we have missed here.
-        if fd < 0 {
+        if fd.get() < 0 {
```
ultimaweapon, line 66: Same here.

ultimaweapon (changes requested):

src/kernel/src/syscalls/output.rs
SuchAFuriousDeath: Sure, ok.

src/kernel/src/process/file.rs
SuchAFuriousDeath: You mean that the check should be in Fd::new?

SuchAFuriousDeath: If so, should this function receive an Fd, or an i32 with Fd::new being called inside the function?

src/kernel/src/fs/dev/dipsw.rs
ultimaweapon: Make it begin with lowercase.

src/kernel/src/process/file.rs
ultimaweapon: Better to accept an `Fd` so the caller know it need to pass a file descriptor.

src/kernel/src/fs/dev/dipsw.rs
SuchAFuriousDeath: Done

src/kernel/src/process/file.rs
SuchAFuriousDeath: Couldn't that diverge too much from what the actual kernel does?

src/kernel/src/fs/mod.rs
```diff
@@ -12,17 +12,61 @@ use param::Param;
...
 mod vnode;
 
+//Represents a file descriptor
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Fd(i32);
+
+impl Fd {
+    pub fn new(fd: i32) -> Self {
+        Self(fd)
+    }
```
ultimaweapon, line 35: This will allow `Fd` to become invalid.

src/kernel/src/fs/mod.rs
```diff
@@ -12,17 +12,61 @@ use param::Param;
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl TryFrom<SysArg> for Fd {
+    type Error = TryFromIntError;
+
+    fn try_from(value: SysArg) -> Result<Self, Self::Error> {
+        Ok(Self(value.try_into()?))
```
ultimaweapon, line 52: This will allow negative value, which is not a valid FD.

src/kernel/src/fs/mod.rs
```diff
@@ -12,17 +12,61 @@ use param::Param;
...
+    }
+}
+
+impl TryFrom<Fd> for usize {
+    type Error = TryFromIntError;
+
+    fn try_from(value: Fd) -> Result<Self, Self::Error> {
+        Ok(value.0.try_into()?)
+    }
+}
```
ultimaweapon, line 62: This one doesn't need to be `TryFrom` if negative value is not possible.

ultimaweapon (changes requested):

src/kernel/src/process/file.rs
SuchAFuriousDeath: I agree, it's better to not have the wrapper in the first place.

ultimaweapon:
> @ultimaweapon I think I have it, but it's not very pretty. This probably requires a thorough review. The slice argument didn't have to be changed right away, but it will have to be changed eventually I think, because of this: image

Do you find any IOCTLs that rely on non-null data while the size is zero? The reason I was designed in this way because I don't see any of those IOCTLs.

SuchAFuriousDeath:
The branch in the screenshot in the comment you responded to seems to imply so.

ultimaweapon:
The reason I don't want to maintain the same behavior is because it is unsafe, which will make it hard for the people who are not familiar with unsafe code to work on. So better to keep it as a slice if no any IOCTLs rely on this. If there are some we need to find a safe solution.

SuchAFuriousDeath:
I think the dipsw ioctl relies in this, with command 0xc0088803.

SuchAFuriousDeath:
And if unsafe code is a concern, perhaps we could make the argument an enum, with one variant being a slice and the ther being in int pointer.

ultimaweapon:
> And if unsafe code is a concern, perhaps we could make the argument an enum, with one variant being a slice and the ther being a reference to an int.

Good idea.

SuchAFuriousDeath:
So, should I do it right away?

ultimaweapon:
If it is not required better to skip it for now so we can merge this faster.

ultimaweapon:
I mean revert back to a slice if you decided to not implement with this PR.

SuchAFuriousDeath:
https://github.com/freebsd/freebsd-src/blob/1da7f3f6f72b2245e458fc7195733268ae4a1136/sys/kern/sys_generic.c#L685-L698
Look at the else branch.

ultimaweapon:
All of the places that use it always accompanied with size, which no-op when the size is zero.

SuchAFuriousDeath:
Oh. interesting

SuchAFuriousDeath:
Now it should be correct, but checks don't pass for some reason.

SuchAFuriousDeath:
Big thanks.

src/kernel/src/fs/mod.rs
```diff
@@ -284,18 +326,29 @@ impl Fs {
...
-            todo!("ioctl with size != 0");
+            if com & IOC_VOID != 0 {
+                todo!("ioctl with com & IOC_VOID != 0");
+            } else {
+                if size > SYS_IOCTL_SMALL_SIZE {
+                    //malloc here
+                    todo!("ioctl with size != 0 and com & IOC_VOID == 0 and size > SYS_IOCTL_SMALL_SIZE")
+                } else {
+                    bytemuck::cast_slice_mut(&mut smalldata)
+                }
```
ultimaweapon, line 340: Where did you get this from? On the 9.0 kernel it will always `malloc` here.

ultimaweapon (changes requested):

src/kernel/src/fs/mod.rs
SuchAFuriousDeath: I don't remember, but you are correct. Sorry

ultimaweapon (approved):
There are some minor issues but the main things are looking good!

PR #447, opened 2023-11-15
Implement syscall 594

SuchAFuriousDeath:


src/kernel/src/rtld/mod.rs
```diff
@@ -431,7 +432,29 @@ impl<E: ExecutionEngine> RuntimeLinker<E> {
...
+    fn sys_dynlib_load_prx(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
+        let libname = unsafe { i.args[0].to_str(1024) }?.unwrap();
+        let _args: usize = i.args[1].into();
+        let p_id: *mut u32 = i.args[2].into();
+
+        if self.app.file_info().is_none() {
+            return Err(SysErr::Raw(EPERM));
+        }
+
+        //TODO implement the rest of this function
```
ultimaweapon, line 444: Why skipping this part?

src/kernel/src/rtld/module.rs
```diff
@@ -16,12 +16,14 @@ use std::io::Write;
 use std::path::Path;
 use std::sync::Arc;
 
+pub type ModuleHandle = u32;
```
ultimaweapon, line 19: Better to use tuple struct instead to achieve strong typing.

ultimaweapon (changes requested):

src/kernel/src/rtld/module.rs
SuchAFuriousDeath: Sure, no problem.

src/kernel/src/rtld/mod.rs
SuchAFuriousDeath: I'm not really sure If I can figure the rest out correctly. This works to a point where we don't immediately crash.

ultimaweapon: Okay. Try your best and I'll take a look at this later.

SuchAFuriousDeath:
@ultimaweapon Is our RuntimeLinker::load supposed to be a full equvalent of load_prx or just partial? Should there be a separate function for load_prx?

ultimaweapon:
The current version is based on the original Obliteration when I have not seen the PS4 kernel yet. Our implementation goal is trying to be close with the PS4 as much as possible. So if you can refactor it then go ahead. If not try to make its behavior the same as PS4.

SuchAFuriousDeath:
The reason I'm asking is that our load function doesn't relocate, but load_prx does relocate.

ultimaweapon:
The reason that `RuntimeLinker::load` do not relocate because it is used to load libkernel and internal libc, which need to relocate together. If you don't want to refactor just create a separate function for load_prx.

SuchAFuriousDeath:
@ultimaweapon Please take over this once you have time, I don't think I can do it in a reasonable amount of time.

ultimaweapon:
Okay. Please close this one and unassign yourself from the issue.

PR #452, opened 2023-11-18
Fixes xsave alignment

ultimaweapon:


VocalFan:
@SuchAFuriousDeath As you had the issue before any of the fixes, try this and see if it still is fixed, if so... Well, LGTM!

SuchAFuriousDeath:
It looks like this doesn't break anything for me

SuchAFuriousDeath (approved):

PR #453, opened 2023-11-18
Fix unused initializations from Sysinfo

### LINKED ISSUE #451: Find a solution to disable sysinfo from spawing rayon thread
Because spawning multiple threads just for this is too much. If it is not possible we need to find the other crate.

VocalFan:
Using New_With_Specifics instead of just New, we now only initialize the CPU and RAM sysinfo, along with disabling the spam of threads. This takes us from 30 threads back to 5 threads!

Fixes #451 

@ultimaweapon

### DIFF
--- a/src/kernel/Cargo.toml
+++ b/src/kernel/Cargo.toml
@@ -23,7 +23,7 @@ param = { path = "../param" }
 serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.9"
 sha1 = "0.10"
-sysinfo = "0.29.10"
+sysinfo = { version = "0.29.10", default-features = false }
 termcolor = "1.2.0"
 thiserror = "1.0"
 tls = { path = "../tls" }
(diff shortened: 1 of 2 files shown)


ultimaweapon:
Wait, it is still spawning the threads?

VocalFan:
> Wait, it is still spawning the threads?

Uh, I just said the total thread the debugger shows when running the kernel. Before, it was 30 total threads. Now there's just 8 threads... Idk how many the kernel usually has.

ultimaweapon:
You can test by setting a breakpoint at https://github.com/obhq/obliteration/blob/main/src/kernel/src/main.rs#L121. Look at the number of threads then step over it to see if it spawn another threads.

VocalFan:
> You can test by setting a breakpoint at https://github.com/obhq/obliteration/blob/main/src/kernel/src/main.rs#L121. Look at the number of threads then step over it to see if it spawn another threads.

Seems like it spawns 3.

*Deleting all sysinfo leaves 5 threads, but adding back sysinfo leaves 8 threads on crash.*

I'm trying to see if Init system with specifics can reduce the thread count.

VocalFan:
ALRIGHT, so, I got it so that we use `System::new_with_specifics` instead of `System::new`

ultimaweapon (approved):

PR #459, opened 2023-11-23
Implement syscall 649

### LINKED ISSUE #435: Unimplemented syscall 649
Reached after stubbing syscall 4. According to the PS4 dev wiki, this one is called 'sys_dynlib_get_obj_member'.

SuchAFuriousDeath:
Should close #435 (for now). I had this one in the drawer for a while, might as well make it official.
I'm not super certain about the unwrap_or(0).

### DIFF
--- a/.github/workflows/ci-mac.yml
+++ b/.github/workflows/ci-mac.yml
@@ -1,42 +1,80 @@
 name: CI (Mac)
 on:
   workflow_call:
+env:
+  CMAKE_BUILD_PARALLEL_LEVEL: '3'
 jobs:
+  llvm:
+    name: LLVM for Mac
+    runs-on: macos-12
+    outputs:
+      dist-key: ${{ steps.cache-keys.outputs.dist }}
+    env:
+      LLVM_URL: https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.5/llvm-project-17.0.5.src.tar.xz
+    steps:
+    - name: Generate cache keys
+      run: |
+        hash=$(md5 -qs "$LLVM_URL")
+        echo "dist=${{ runner.os }}-llvm-$hash" >> $GITHUB_OUTPUT
+      id: cache-keys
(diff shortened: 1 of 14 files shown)


SuchAFuriousDeath:
@ultimaweapon By the way do you think we should make vscode run fmt on Ctrl+S? I think it can be done fairly easily.
https://stackoverflow.com/questions/67859926/how-to-run-cargo-fmt-on-save-in-vscode

SuchAFuriousDeath:
Okay, this is wrong. It's not proc_param, it's module param :(. Close enough.

SuchAFuriousDeath:
@ultimaweapon This is probably wrong, but close to being correct. Hopefully you can finish it off.

ultimaweapon:
> @ultimaweapon By the way do you think we should make vscode run fmt on Ctrl+S? I think it can be done fairly easily.
> https://stackoverflow.com/questions/67859926/how-to-run-cargo-fmt-on-save-in-vscode

Good idea.

SuchAFuriousDeath:
Actually, I think this is correct after all.

SuchAFuriousDeath:
Just for clarity, though: I think this is logically correct, but I think this working depends on a correct implementation of Syscall 549. co we can't know for sure until that is done.

ultimaweapon:
What are you looking for from 594?

SuchAFuriousDeath:
I'm not actually sure. It just threw a segfault with my partial 594 implementation.

SuchAFuriousDeath:
@ultimaweapon WIth your 594, this does reach tsc_freq. We still have a segfault after that for unknown reasons.

ultimaweapon:
> @ultimaweapon WIth your 594, this does reach tsc_freq. (Which it didn't before) We still have a segfault after that for unknown reasons.

Let's do what we can do for now, as accurate as possible.

ultimaweapon (approved):

PR #462, opened 2023-11-25
Implements syscall 594

### LINKED ISSUE #428: Unimplemented syscall 594

ultimaweapon:
Closes #428.

### DIFF
--- a/src/kernel/src/fs/mod.rs
+++ b/src/kernel/src/fs/mod.rs
@@ -1,14 +1,12 @@
 pub use self::file::*;
 pub use self::item::*;
 pub use self::path::*;
-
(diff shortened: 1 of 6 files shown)


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Great work!

PR #472, opened 2023-11-26
Implement machdep_tsc_freq

SuchAFuriousDeath:


SuchAFuriousDeath:
Maybe the linker error is completely unrelated to the assembly, I'll test tomorrow.

SuchAFuriousDeath:
I was just writing a comment asking you what I should do with the aarch64 'branch'.

src/kernel/src/arch/mod.rs
```diff
@@ -51,4 +57,52 @@ impl MachDep {
...
+
+        while Instant::now().duration_since(start) < duration {
+            // Busy-wait
+        }
+
+        let end_tsc = unsafe { std::arch::x86_64::_rdtsc() };
+        let elapsed = end_tsc - start_tsc;
+
+        AtomicU64::new(elapsed * 1000)
+    }
```
ultimaweapon, line 80: Did you take this from the PS4 or FreeBSD? What I saw in the PS4 is not like this.

src/kernel/src/arch/mod.rs
```diff
@@ -51,4 +57,52 @@ impl MachDep {
...
+
+        unsafe {
+            asm!("mrs {0}, cntvct_el0", out(reg) end_tsc);
+        }
+
+        // Calculate the elapsed time
+        let elapsed = end_tsc - start_tsc;
+
+        AtomicU64::new(elapsed * 1000)
+    }
```
ultimaweapon, line 107: Have you tested this? If not better to put `todo!` for non-x86 here instead.

ultimaweapon (changes requested):

src/kernel/src/arch/mod.rs
SuchAFuriousDeath: Ok, will do.

src/kernel/src/arch/mod.rs
SuchAFuriousDeath: FreeBSD.

src/kernel/src/sysctl/mod.rs
```diff
@@ -822,3 +879,35 @@ static HW_PAGESIZE: Oid = Oid {
...
+
+static MACHDEP_CHILDREN: OidList = OidList {
+    first: Some(&MACHDEP_TSC_FREQ), // TODO: Use a proper value.
+};
+
+static MACHDEP_TSC_FREQ: Oid = Oid {
+    parent: &MACHDEP_CHILDREN,
+    link: None, // TODO: Implement this.
+    number: Sysctl::MACHDEP_TSC_FREQ,
+    kind: Sysctl::CTLFLAG_RD | Sysctl::CTLTYPE_U64,
```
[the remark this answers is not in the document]
SuchAFuriousDeath: Fixed.

src/kernel/src/sysctl/mod.rs
```diff
@@ -822,3 +879,35 @@ static HW_PAGESIZE: Oid = Oid {
...
+    link: None, // TODO: Implement this.
+    number: Sysctl::CTL_MACHDEP,
+    kind: Sysctl::CTLFLAG_RW | Sysctl::CTLTYPE_NODE,
+    arg1: Some(&MACHDEP_CHILDREN),
+    arg2: 0,
+    name: "machdep",
+    handler: None,
+    fmt: "N",
+    descr: "machine dependent",
+    enabled: true,
```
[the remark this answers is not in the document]
SuchAFuriousDeath: fixed

src/kernel/src/arch/mod.rs
ultimaweapon: Seems like we have a complicated problem here. The problem is `rdtsc` is not really usablec here because the current thread might get suspended anytime. The best possible solution here should be query this value from the OS.

ultimaweapon: We might need to hard-coded the value here for now. From what I know the only use case for this value is timer system. That mean the worst case for wrong value here is inaccurate timer.

SuchAFuriousDeath: So, what should I do with this PR?

SuchAFuriousDeath: Ok.

src/kernel/src/sysctl/mod.rs
```diff
@@ -822,3 +879,35 @@ static HW_PAGESIZE: Oid = Oid {
...
+    link: None, // TODO: Implement this.
+    number: Sysctl::MACHDEP_TSC_FREQ,
+    kind: Sysctl::CTLFLAG_RW | Sysctl::CTLTYPE_U64,
+    arg1: None,
+    arg2: 0,
+    name: "tsc_freq",
+    handler: Some(Sysctl::machdep_tsc_freq),
+    fmt: "QU",
+    descr: "Time Stamp Counter frequency",
+    enabled: false,
```
ultimaweapon, line 912: This field is 1 on the PS4. Actually you was fixed the wrong one.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #491, opened 2023-12-03
Implements syscall 464

### LINKED ISSUE #490: Unimplemented syscall 464
According to one user of our discord, this is reached by God of War.
This seems to be the FreeBSD syscall thr_set_name.

SuchAFuriousDeath:
Will close #490.

### DIFF
--- a/src/kernel/src/process/mod.rs
+++ b/src/kernel/src/process/mod.rs
@@ -417,17 +417,17 @@ impl VProc {
     }
 
     fn sys_thr_set_name(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
-        let name: Option<&str> = unsafe { i.args[0].to_str(32) }?;
-        let id: i32 = i.args[1].try_into().unwrap();
+        let tid: i32 = i.args[0].try_into().unwrap();
+        let name: Option<&str> = unsafe { i.args[1].to_str(32) }?;
 
-        if id == -1 {
+        if tid == -1 {
             self.set_name(name);
         } else {
             let threads = self.threads.read();
 
             let thr = threads
                 .iter()
-                .find(|t| t.id().get() == id)
+                .find(|t| t.id().get() == tid)
                 .ok_or(SysErr::Raw(ESRCH))?;
 
             thr.set_name(name);
--- a/src/kernel/src/process/mod.rs
+++ b/src/kernel/src/process/mod.rs
@@ -49,7 +49,7 @@ pub struct VProc {
     sigacts: Gutex<SignalActs>,                      // p_sigacts
     files: FileDesc,                                 // p_fd
     limits: [ResourceLimit; ResourceLimit::NLIMITS], // p_limit
-    comm: Gutex<String>,                             // p_comm
+    comm: Gutex<Option<String>>,                     // p_comm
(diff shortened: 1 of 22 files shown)


SuchAFuriousDeath:
@ultimaweapon It seems that one of your refactors broke the Mac build, because some used library features are unstable there. At least that's how I understand it.

SuchAFuriousDeath:
Ok, that's weird, rerunning the build made it pass. Whatever.

ultimaweapon:
> Not gonna lie, I'm a bit confused by the comment above the VProc struct. @ultimaweapon What will we do if a game calls fork? Do we expect no game to ever call fork? Is the 'threads' field of 'VProc' an equivalent of td_hash?
> 
> The reason I'm asking is, I need to know what the equivalent of tdfind is for us.

The game should never call to fork. IIRC tdfind is just a loop finding a thread with specific ID, which you can use Rust iterator to achieve the same function.

ultimaweapon:
Seems like macOS randomly use Rust 1.73.

src/kernel/src/process/mod.rs
```diff
@@ -409,6 +416,26 @@ impl VProc {
         Ok(SysOut::ZERO)
     }
 
+    fn sys_thr_set_name(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
+        let name: &str = unsafe { i.args[0].to_str(32) }?.unwrap();
```
ultimaweapon, line 420: The PS4 will use an empty name if this argument is null.

ultimaweapon (changes requested):

src/kernel/src/process/mod.rs
SuchAFuriousDeath: Fixed., thanks.

src/kernel/src/process/mod.rs
```diff
@@ -409,6 +416,26 @@ impl VProc {
         Ok(SysOut::ZERO)
     }
 
+    fn sys_thr_set_name(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
+        let name: Option<&str> = unsafe { i.args[0].to_str(32) }?;
```
ultimaweapon, line 420: This argument is not a name.

src/kernel/src/process/mod.rs
```diff
@@ -409,6 +416,26 @@ impl VProc {
...
+        let id: i32 = i.args[1].try_into().unwrap();
```
ultimaweapon, line 421: This argument is not an ID.

src/kernel/src/process/mod.rs
```diff
@@ -74,6 +75,7 @@ impl VProc {
             dmem_container: gg.spawn(0), // TODO: Check the initial value on the PS4.
             budget: gg.spawn(None),
             limits,
+            comm: gg.spawn(None), //Find out how this is actually set
```
ultimaweapon, line 78: Better to prefix with `TODO:` so it is easy to search for what need to be done.

ultimaweapon (changes requested):

SuchAFuriousDeath:
Good catch. I don't know how I coud get that wrong :(.

ultimaweapon (approved):

PR #504, opened 2023-12-07
Fix/Add Freqs

VocalFan:
Fixes the frequency we currently use.

*There was an extra 0, so we basically told it the CPU was at 16GHz instead of 1.6GHz*

Along with that, I also put in comments for later the PS4 Pro/Neo frequency.


VocalFan:
@SuchAFuriousDeath Done!

SuchAFuriousDeath (approved):

PR #506, opened 2023-12-09
Refactor ioctl com

SuchAFuriousDeath:
Provides a wrapper type for the ioctl argument 'com' along with a couple of macros borrowed from FreeBSD for defining commands.

### DIFF
--- a/src/kernel/src/fs/file.rs
+++ b/src/kernel/src/fs/file.rs
@@ -103,16 +103,6 @@ impl IoctlCom {
         Some(Self(com))
     }
 
-    const fn new(inout: u32, group: u8, num: u8, len: usize) -> Self {
-        let len: u32 = if len > (u32::MAX) as usize {
-            panic!("IOCPARM_LEN is too large");
-        } else {
-            len as u32
-        };
-
-        Self(inout | ((len & Self::IOCPARM_MASK) << 16) | ((group as u32) << 8) | (num as u32))
-    }
-
     pub fn size(&self) -> usize {
         Self::iocparm_len(self.0)
     }
@@ -142,7 +132,13 @@ impl IoctlCom {
     }
 
     pub const fn ioc(inout: u32, group: u8, num: u8, len: usize) -> Self {
-        Self::new(inout, group, num, len)
+        let len: u32 = if len > (u32::MAX) as usize {
+            panic!("IOCPARM_LEN is too large");
+        } else {
+            len as u32
+        };
+
+        Self(inout | ((len & Self::IOCPARM_MASK) << 16) | ((group as u32) << 8) | (num as u32))
     }
 
     pub const fn io(group: u8, num: u8) -> Self {
--- a/src/kernel/src/fs/file.rs
+++ b/src/kernel/src/fs/file.rs
@@ -67,11 +67,124 @@ pub trait VFileOps: Debug + Send + Sync + Display {
     fn ioctl(
         &self,
         file: &VFile,
-        com: u64,
+        com: IoctlCom,
         data: &mut [u8],
         cred: &Ucred,
         td: &VThread,
     ) -> Result<(), Box<dyn Errno>>;
+}
+
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct IoctlCom(u32);
+
+impl Display for IoctlCom {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "0x{:x}", self.0)
+    }
+}
+
+impl IoctlCom {
+    pub const IOCPARM_SHIFT: u32 = 13;
+    pub const IOCPARM_MASK: u32 = (1 << Self::IOCPARM_SHIFT) - 1;
+    pub const IOC_VOID: u32 = 0x20000000;
+    pub const IOC_OUT: u32 = 0x40000000;
+    pub const IOC_IN: u32 = 0x80000000;
+    pub const IOC_INOUT: u32 = Self::IOC_IN | Self::IOC_OUT;
+
+    pub const fn new_truncated(com: u64) -> Self {
+        Self(com as u32)
+    }
+
+    pub const fn new_exact(com: u32) -> Self {
+        Self(com)
+    }
+
+    pub fn size(&self) -> usize {
+        ((self.0 >> 16) & Self::IOCPARM_MASK) as usize
+    }
+
+    pub fn is_void(&self) -> bool {
+        self.0 & Self::IOC_VOID != 0
+    }
+
+    pub fn is_out(&self) -> bool {
+        self.0 & Self::IOC_OUT != 0
+    }
+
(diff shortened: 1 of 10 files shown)


src/kernel/src/fs/file.rs
ultimaweapon, line 1: Better to use `const fn` instead of the macros if possible. The reason is because it is much more easier to read.

ultimaweapon (changes requested):

src/kernel/src/fs/file.rs
SuchAFuriousDeath: Sure, ok. Didn't know you were still awake. By the way, the dmem ioctl is extremely easy. PR coming in a minute.

SuchAFuriousDeath: Wasn't so easy after all :(

SuchAFuriousDeath:
@ultimaweapon Refactored it to const functions.

src/kernel/src/fs/file.rs
```diff
@@ -67,13 +68,104 @@ pub trait VFileOps: Debug + Send + Sync + Display {
...
+                && Self::iocparm_len(com) != 4)
+    }
+
+    pub const fn iocparm_len(com: u32) -> usize {
+        ((com >> 16) & Self::IOCPARM_MASK) as usize
+    }
+
+    pub const fn ioc(inout: u32, group: u8, num: u8, len: usize) -> Self {
+        Self::new(inout, group, num, len)
+    }
```
ultimaweapon, line 146: Better to remove this and use `new` directly so the code will be easier to follow.

src/kernel/src/fs/file.rs
```diff
@@ -67,13 +68,104 @@ pub trait VFileOps: Debug + Send + Sync + Display {
...
+
+    pub const fn io(group: u8, num: u8) -> Self {
+        Self::ioc(Self::IOC_VOID, group, num, 0)
+    }
+
+    pub const fn iowint(group: u8, num: u8) -> Self {
+        Self::ioc(Self::IOC_IN, group, num, std::mem::size_of::<i32>())
```
ultimaweapon, line 153: Is not this one `IOC_VOID`?

src/kernel/src/fs/file.rs
```diff
@@ -67,13 +68,104 @@ pub trait VFileOps: Debug + Send + Sync + Display {
...
+    pub fn is_in(&self) -> bool {
+        self.0 & Self::IOC_IN != 0
+    }
+
+    pub const fn is_invalid(com: u32) -> bool {
+        (com & (Self::IOC_VOID | Self::IOC_IN | Self::IOC_OUT) == 0)
+            || (com & (Self::IOC_IN | Self::IOC_OUT)) != 0 && Self::iocparm_len(com) == 0
+            || (com & Self::IOC_VOID != 0
+                && Self::iocparm_len(com) != 0
+                && Self::iocparm_len(com) != 4)
```
ultimaweapon, line 137: Is it possible to split the whole this into a separated `if` so it will be more easier to read?

src/kernel/src/fs/file.rs
```diff
@@ -67,13 +68,104 @@ pub trait VFileOps: Debug + Send + Sync + Display {
...
+            panic!("IOCPARM_LEN is too large");
+        } else {
+            len as u32
+        };
+
+        Self(inout | ((len & Self::IOCPARM_MASK) << 16) | ((group as u32) << 8) | (num as u32))
+    }
+
+    pub fn size(&self) -> usize {
+        Self::iocparm_len(self.0)
```
ultimaweapon, line 117: Better to inline `iocparm_len` here then remove it so the code is easier to follow.

src/kernel/src/fs/file.rs
```diff
@@ -67,13 +68,104 @@ pub trait VFileOps: Debug + Send + Sync + Display {
...
+
+impl IoctlCom {
+    pub const IOCPARM_SHIFT: u32 = 13;
+    pub const IOCPARM_MASK: u32 = (1 << Self::IOCPARM_SHIFT) - 1;
+    pub const IOC_VOID: u32 = 0x20000000;
+    pub const IOC_OUT: u32 = 0x40000000;
+    pub const IOC_IN: u32 = 0x80000000;
+    pub const IOC_INOUT: u32 = Self::IOC_IN | Self::IOC_OUT;
+
+    pub const fn try_from_raw(com: u64) -> Result<Self, SysErr> {
```
ultimaweapon, line 96: Better to return `Option` and let the caller decide what to do instead.

ultimaweapon (changes requested):

src/kernel/src/fs/file.rs
SuchAFuriousDeath: Good idea.

src/kernel/src/fs/file.rs
SuchAFuriousDeath: It is

SuchAFuriousDeath:
@ultimaweapon Everything should be fixed now.

src/kernel/src/fs/file.rs
ultimaweapon: @SuchAFuriousDeath ops sorry. I just saw the `iocparm_len` version accept the raw value, not the `self`. So your first version here is the correct way to go.

SuchAFuriousDeath: Oh :D

SuchAFuriousDeath: @ultimaweapon reverted

SuchAFuriousDeath:
By the way in the meantime, I looked at sys_revoke. One of the functions called there is vn_printf. inside that function, one of the pointers to static data points to a pointer to this: 
image

SuchAFuriousDeath:
Also, is it ok now? :D

SuchAFuriousDeath:
@ultimaweapon By the way, I don't really remember how I got to vn_printf, but I'm pretty sure you can find it by searching for strings in the FreeBSD source code and finding the functions that references them.

ultimaweapon (changes requested):
Looking good! Just one remaining issue.

src/kernel/src/syscalls/input.rs
```diff
@@ -126,3 +126,11 @@ impl TryFrom<SysArg> for u8 {
...
     }
 }
+
+impl TryFrom<SysArg> for IoctlCom {
+    type Error = SysErr;
+
+    fn try_from(v: SysArg) -> Result<IoctlCom, Self::Error> {
+        IoctlCom::try_from_raw(v.into()).ok_or(SysErr::Raw(ENOTTY))
+    }
+}
```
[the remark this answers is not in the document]
SuchAFuriousDeath: Yeah, you already told me this exact thing, I just somehow flipped it around in my head.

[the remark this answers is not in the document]
SuchAFuriousDeath: Fixed.

ultimaweapon (approved):

PR #507, opened 2023-12-09
Initializes FS rework

ultimaweapon:
So our FS will be similar to the PS4.


VocalFan (approved):
Seems to match the Filesystem with our new FS knowledge, nice!

PR #512, opened 2023-12-09
Implements dmem1 ioctl

### LINKED ISSUE #503: Implement dmem ioctl
Reached after #500

SuchAFuriousDeath:
Fixes #503

### DIFF
--- a/src/kernel/src/fs/dev/dipsw.rs
+++ b/src/kernel/src/fs/dev/dipsw.rs
@@ -69,7 +69,7 @@ impl Display for Dipsw {
 
 #[derive(Error, Debug)]
 enum IoctlErr {
-    #[error("invalid command passed to Dipsw::ioctl")]
+    #[error("invalid command passed to dipsw ioctl")]
     InvalidCommand,
 }
 
--- a/src/kernel/src/fs/dev/dipsw.rs
+++ b/src/kernel/src/fs/dev/dipsw.rs
@@ -14,21 +14,21 @@ pub struct Dipsw {}
 impl Dipsw {
     pub const PATH: &VPath = vpath!("/dev/dipsw");
 
+    const COM1: IoctlCom = IoctlCom::iow::<i16>(0x88, 1); //TODO: figure out actual type
+    const COM2: IoctlCom = IoctlCom::iow::<i16>(0x88, 2); //TODO: figure out actual type
+    const COM3: IoctlCom = IoctlCom::iowr::<i64>(0x88, 3); //TODO: figure out actual type
+    const COM4: IoctlCom = IoctlCom::iow::<(i64, i64)>(0x88, 4); //TODO: figure out actual type, probably a struct
+    const COM5: IoctlCom = IoctlCom::iow::<(i64, i64)>(0x88, 5); //TODO: figure out actual type, probably a struct
+    const COM6: IoctlCom = IoctlCom::ior::<i32>(0x88, 6);
+    const COM7: IoctlCom = IoctlCom::ior::<i32>(0x88, 7); //TODO: figure out actual type
+    const COM8: IoctlCom = IoctlCom::ior::<i64>(0x88, 8); //TODO: figure out actual type
+    const COM9: IoctlCom = IoctlCom::ior::<i64>(0x88, 9); //TODO: figure out actual type
+    const COM10: IoctlCom = IoctlCom::iow::<(i64, i64)>(0x88, 10); //TODO: figure out actual type, probably a struct
+
     pub fn new() -> Self {
         Self {}
     }
 }
-
-const COM1: IoctlCom = IoctlCom::iow::<i16>(0x88, 1); //TODO: figure out actual type
-const COM2: IoctlCom = IoctlCom::iow::<i16>(0x88, 2); //TODO: figure out actual type
-const COM3: IoctlCom = IoctlCom::iowr::<i64>(0x88, 3); //TODO: figure out actual type
-const COM4: IoctlCom = IoctlCom::iow::<(i64, i64)>(0x88, 4); //TODO: figure out actual type, probably a struct
-const COM5: IoctlCom = IoctlCom::iow::<(i64, i64)>(0x88, 5); //TODO: figure out actual type, probably a struct
-const COM6: IoctlCom = IoctlCom::ior::<i32>(0x88, 6);
-const COM7: IoctlCom = IoctlCom::ior::<i32>(0x88, 7); //TODO: figure out actual type
-const COM8: IoctlCom = IoctlCom::ior::<i64>(0x88, 8); //TODO: figure out actual type
-const COM9: IoctlCom = IoctlCom::ior::<i64>(0x88, 9); //TODO: figure out actual type
-const COM10: IoctlCom = IoctlCom::iow::<(i64, i64)>(0x88, 10); //TODO: figure out actual type, probably a struct
 
 impl VFileOps for Dipsw {
     fn write(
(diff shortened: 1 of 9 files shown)


src/kernel/src/fs/dev/dmem1.rs
```diff
@@ -24,12 +28,20 @@ impl VFileOps for Dmem1 {
...
+        com: u64,
+        mut data: &mut [u8],
         _: &Ucred,
         _: &VThread,
     ) -> Result<(), Box<dyn Errno>> {
-        todo!()
+        match com {
+            Self::COM10 => {
+                data.write_u64::<LittleEndian>(Self::TOTAL_SIZE as u64)
+                    .unwrap();
```
ultimaweapon, line 39: Better to write as a native endian instead unless the caller expect little-endian.

src/kernel/src/fs/dev/dmem1.rs
```diff
@@ -24,12 +28,20 @@ impl VFileOps for Dmem1 {
...
         _: &crate::fs::VFile,
-        _: u64,
-        _: &mut [u8],
+        com: u64,
+        mut data: &mut [u8],
         _: &Ucred,
         _: &VThread,
     ) -> Result<(), Box<dyn Errno>> {
-        todo!()
+        match com {
```
ultimaweapon, line 36: Why skip the 3 checks before this?

ultimaweapon (changes requested):

src/kernel/src/fs/dev/dmem1.rs
SuchAFuriousDeath: I forgot because I was too excited :( I'm looking at this now and wondering. On the PS4, the auth field of ucred starts at offset 0x58, right? Otherwise, nothing makes sense to me.

ultimaweapon: Correct.

SuchAFuriousDeath:
@ultimaweapon I think I have the checks, but the ioctl fails. So either I made a mistake or the authinfo is wrong.

ultimaweapon:
The auth info for the main thread is sys core, which should be the most privileged one.

SuchAFuriousDeath:
https://github.com/obhq/obliteration/blob/65761d9342a9685a6fcd7dbbeab30683134ada24/src/kernel/src/fs/file.rs#L132-L146

SuchAFuriousDeath:
I don't know if If my memory is bad or something changed in main in the meantime , but there's now a dipsw ioctl (an unimplemented one) before we reach the dmem ioctl.

ultimaweapon:
Might be caused by https://github.com/obhq/obliteration/pull/505

SuchAFuriousDeath:
That's probably it.

SuchAFuriousDeath:
Ok, that's not actually it. And I'm extremely confused. One might notice that I moved the COM* constants into the Dipsw structs, but I haven't adjusted the match statement. What's extremely weird is that the code actually compiles, but somehow, the match statement matches on COM1 instead of COM6. This is extremely weird, I don't think it should compile.

SuchAFuriousDeath:
@ultimaweapon It should all be fixed now.

src/kernel/src/fs/dev/dipsw.rs
```diff
@@ -46,25 +47,25 @@ impl VFileOps for Dipsw {
...
+            Self::COM1 => todo!("dipsw ioctl 0x80028801"),
+            Self::COM2 => todo!("dipsw ioctl 0x80028802"),
+            Self::COM3 => todo!("dipsw ioctl 0xc0088803"),
+            Self::COM4 => todo!("dipsw ioctl 0x80108804"),
+            Self::COM5 => todo!("dipsw ioctl 0x80108805"),
+            Self::COM6 => {
                 //todo write the correct value if unk_func1() = false and
                 // unk_func2() = true
-                data.write_i32::<LittleEndian>(false as i32).unwrap();
+                data.copy_from_slice(&(false as i32).to_le_bytes());
```
ultimaweapon, line 63: Does the PS4 expect this value to be native endian? If yes better to write this as a native endian so in the future when we ported to big-endian it will be working correctly.

src/kernel/src/ucred/mod.rs
```diff
@@ -52,6 +52,14 @@ impl Ucred {
         self.auth.caps.is_system()
     }
 
+    pub fn is_unk1(&self) -> bool {
+        self.auth.caps.is_unk1() || self.auth.attrs.is_unk1()
```
ultimaweapon, line 56: I think this one required both condition to be true instead of either one is true.

ultimaweapon (changes requested):

src/kernel/src/fs/dev/dipsw.rs
SuchAFuriousDeath: So all the other times we call to_le_bytes, it should be to_ne bytes too?

ultimaweapon: If the PS4 treat it as a native when reading or writing then yes.

SuchAFuriousDeath: Ok

src/kernel/src/ucred/mod.rs
```diff
@@ -52,6 +52,14 @@ impl Ucred {
         self.auth.caps.is_system()
     }
 
+    pub fn is_unk1(&self) -> bool {
+        self.auth.caps.is_unk1() && self.auth.attrs.is_unk1()
+    }
+
+    pub fn is_unk2(&self) -> bool {
+        self.auth.caps.is_unk1() || self.auth.attrs.is_unk2()
```
ultimaweapon, line 60: I think this one is also the same as `is_unk1`.

ultimaweapon (changes requested):

src/kernel/src/fs/dev/dmem1.rs
```diff
@@ -1,18 +1,33 @@
...
-    pub fn new() -> Self {
-        Self {}
+    pub const DMEM_GRP: u8 = 0x80;
+
+    pub const COM10: IoctlCom = IoctlCom::ior::<usize>(Self::DMEM_GRP, 0xa);
+
+    pub fn new(vp: &Arc<VProc>) -> Self {
+        Self {
+            vp: vp.clone(),
+            total_size: 0x180_000_000, // 6 GB
```
ultimaweapon, line 28: Where did you get this number from? What I see on the PS4 is it is capped at `0x13C000000`.

ultimaweapon (changes requested):

src/kernel/src/fs/dev/dmem1.rs
SuchAFuriousDeath: I got it from fpps4, as I don't know how to get it from the actual PS4

ultimaweapon: Screenshot from 2023-12-16 16-55-52

SuchAFuriousDeath: It's all zeroed for me. Is it initialized somewhere else?

ultimaweapon: Strange. If what you have is a memory dump of the kernel and you load it to Ghidra as a raw memory with correct address it should not be zeroed.

SuchAFuriousDeath: I loaded it using the orbis plugin.

SuchAFuriousDeath: So, what should I do? Return the 0x13C000000?

ultimaweapon: The correct implementation here is use the value from this dmem device. If you don't know this value then you can use `0x13C000000` with `TODO`.

SuchAFuriousDeath: I changed it.

ultimaweapon (approved):

PR #514, opened 2023-12-10
Changes to PKG installer

### LINKED ISSUE #283: Allow installation of update PKGs.
So, I made fPKGs of the disk copy that I have of Tom Clancy The Division along with the patch. I installed the Base PKG, then I went to install the Patch PKG, but it seems that we don't check if a PKG is a patch yet seeing as I just get 

image

Maybe we can check metadata or file names?
Base name: UP0001-CUSA01810_00-THEDIVISION00001-A0100-V0100.pkg
Update name: UP0001-CUSA01810_00-THEDIVISION00001-A0121-V0100.pkg

VocalFan:
Fixes #283 

Implements it so that Patch and DLC pkgs are supported

Patches are put under a CUSA-PATCH-VERSION folder while DLC is ready for later implementation
image image

Adds App_ver, Category, Content_ID and Version to the Param.sfo reader. (Technically, there is also a variant called shortContent_ID, which is just Content_ID except only the last part, which is what the PS4 uses to list DLCs.)
image image


VocalFan:
@ultimaweapon There we go :P

ultimaweapon (approved):

PR #523, opened 2023-12-16
Implement syscall 653

SuchAFuriousDeath:


ultimaweapon (changes requested):
From what I saw block pool system is a part of dmem so better to put it there.

SuchAFuriousDeath:
Sure, will do.

ultimaweapon (approved):

PR #534, opened 2023-12-23
Partially implements console ioctl with com = 0x20007461

### LINKED ISSUE #525: Implement console ioctl with com = 0x20007461
Now that we know what it is, it should be fairly easy.

SuchAFuriousDeath:
Progresses #525.

### DIFF
--- a/src/kernel/src/fs/dev/console.rs
+++ b/src/kernel/src/fs/dev/console.rs
@@ -1,4 +1,4 @@
-use crate::errno::{Errno, ENXIO, EPERM};
+use crate::errno::{Errno, ENXIO};
 use crate::fs::{IoctlCom, VFile, VFileOps, VPath};
 use crate::process::{VProc, VThread};
 use crate::tty::Tty;
@@ -13,7 +13,7 @@ use thiserror::Error;
 /// An implementation of `/dev/console`.
 #[derive(Debug)]
 pub struct Console {
-    tty: Tty,
+    tty: Arc<Tty>,
 }
 
 impl Console {
--- a/src/kernel/src/fs/dev/console.rs
+++ b/src/kernel/src/fs/dev/console.rs
@@ -1,23 +1,31 @@
-use crate::errno::Errno;
+use crate::errno::{Errno, ENXIO};
 use crate::fs::{IoctlCom, VFile, VFileOps, VPath};
-use crate::process::VThread;
+use crate::process::{VProc, VProcGroup, VThread};
 use crate::ucred::Ucred;
 use macros::vpath;
 use std::fmt::{Display, Formatter};
 use std::io::{self, Write};
+use std::num::NonZeroI32;
+use std::sync::Arc;
+use thiserror::Error;
 
 /// An implementation of `/dev/console`.
 #[derive(Debug)]
-pub struct Console {}
+pub struct Console {
+    tty: Tty,
+}
 
 impl Console {
     pub const PATH: &VPath = vpath!("/dev/console");
 
-    #[allow(dead_code)]
-    pub const TIOCSCTTY: IoctlCom = IoctlCom::io(b't', 97);
+    pub const TTY_GRP: u8 = b't';
+    pub const TIOCSCTTY: IoctlCom = IoctlCom::io(Self::TTY_GRP, 97);
+    pub const TIOCSTI: IoctlCom = IoctlCom::iow::<u8>(Self::TTY_GRP, 114);
 
-    pub fn new() -> Self {
-        Self {}
+    pub fn new(vp: &Arc<VProc>) -> Self {
+        Self {
+            tty: Tty::new(vp.clone()),
+        }
     }
 }
 
@@ -41,17 +49,78 @@ impl VFileOps for Console {
     fn ioctl(
         &self,
         _file: &VFile,
-        _com: IoctlCom,
+        com: IoctlCom,
         _data: &mut [u8],
         _cred: &Ucred,
         _td: &VThread,
     ) -> Result<(), Box<dyn Errno>> {
-        // TODO: Implement this.
+        if self.tty.is_gone() || !self.tty.is_open() {
+            return Err(Box::new(IoctlErr::TtyNotAvailable));
+        }
+
+        match com {
+            Self::TIOCSCTTY => {
+                todo!();
+            }
+            _ => todo!("Unimplemented console ioctl command: {com:?}"),
+        }
+
         Ok(())
     }
 }
 impl Display for Console {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         Self::PATH.fmt(f)
+    }
+}
+
+#[derive(Debug, Error)]
+enum IoctlErr {
+    #[error("tty is not available")]
+    TtyNotAvailable,
+}
+
+impl Errno for IoctlErr {
+    fn errno(&self) -> NonZeroI32 {
+        match self {
+            Self::TtyNotAvailable => ENXIO,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Tty {
+    vp: Arc<VProc>,
+    group: Arc<Option<VProcGroup>>,
+}
+
+impl Tty {
+    pub fn new(vp: Arc<VProc>) -> Self {
+        Self {
+            vp,
+            group: Arc::new(None),
+        }
+    }
+
+    //TODO: implement this
+    pub fn is_gone(&self) -> bool {
+        false
+    }
+
+    //TODO: implement this
+    pub fn is_open(&self) -> bool {
+        true
+    }
+
+    pub fn ioctl(
+        &self,
+        _com: IoctlCom,
+        _data: &mut [u8],
+        _td: &VThread,
+    ) -> Result<(), Box<dyn Errno>> {
+        let grp = self.vp.group().expect("Process group is not set.");
+        let session = grp.session();
+
+        Ok(())
     }
 }
--- a/src/kernel/src/fs/dev/console.rs
+++ b/src/kernel/src/fs/dev/console.rs
@@ -47,17 +47,17 @@ impl VFileOps for Console {
         &self,
         _file: &VFile,
         com: IoctlCom,
-        _data: &mut [u8],
+        data: &mut [u8],
         _cred: &Ucred,
         td: &VThread,
     ) -> Result<(), Box<dyn Errno>> {
         if self.tty.is_gone() || !self.tty.is_open() {
-            return Err(Box::new(IoctlErr::TtyNotAvailable));
+            return Err(Box::new(IoctlError::TtyNotAvailable));
         }
 
         //TODO: implement tty_wait_background and the rest of the checks here.
 
-        self.tty.ioctl(com, _data, td)?;
+        self.tty.ioctl(com, data, td)?;
 
         Ok(())
     }
@@ -69,12 +69,12 @@ impl Display for Console {
 }
 
 #[derive(Debug, Error)]
-enum IoctlErr {
+enum IoctlError {
     #[error("tty is not available")]
     TtyNotAvailable,
 }
 
-impl Errno for IoctlErr {
+impl Errno for IoctlError {
     fn errno(&self) -> NonZeroI32 {
         match self {
             Self::TtyNotAvailable => ENXIO,
--- a/src/kernel/src/fs/dev/console.rs
+++ b/src/kernel/src/fs/dev/console.rs
@@ -1,6 +1,7 @@
-use crate::errno::{Errno, ENXIO};
+use crate::errno::{Errno, ENXIO, EPERM};
 use crate::fs::{IoctlCom, VFile, VFileOps, VPath};
-use crate::process::{VProc, VProcGroup, VThread};
+use crate::process::{VProc, VThread};
+use crate::tty::Tty;
 use crate::ucred::Ucred;
 use macros::vpath;
 use std::fmt::{Display, Formatter};
@@ -17,10 +18,6 @@ pub struct Console {
 
 impl Console {
     pub const PATH: &VPath = vpath!("/dev/console");
-
-    pub const TTY_GRP: u8 = b't';
-    pub const TIOCSCTTY: IoctlCom = IoctlCom::io(Self::TTY_GRP, 97);
-    pub const TIOCSTI: IoctlCom = IoctlCom::iow::<u8>(Self::TTY_GRP, 114);
 
     pub fn new(vp: &Arc<VProc>) -> Self {
         Self {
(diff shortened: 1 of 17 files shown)


SuchAFuriousDeath:
Thank god for Option::is_some_and.

SuchAFuriousDeath:
By the way, I'm pretty sure I found the proc0_init function.

SuchAFuriousDeath:
It seems that kern.proc.appinfo is connected to workaround and to something with cpusets

ultimaweapon:
I can't take a look due to COVID-19. The headache is so bad even though I am already vaccinated.

SuchAFuriousDeath:
It's fine, take care. 

This looks like it initializes budgets for processes.
image

SuchAFuriousDeath:
> It seems that kern.proc.appinfo is connected to workaround and to something with cpusets

Ok, I uncovered it now. The something to do with cpusets just means it calls 'pfind'. I didn't know what the function was up until now :D

SuchAFuriousDeath:
image
This is the initial budget for a process.

SuchAFuriousDeath:
I'm not sure if the fields are named correctly, but their offsets are 0xaf0 and 0xaf4.

SuchAFuriousDeath:
There's also eventhandlers for id tables 
image

SuchAFuriousDeath:
Looking at this now, this probably leaks memory because of a reference cycle.

SuchAFuriousDeath:
By the way, do all the components have to be in an Arc? Couldn't we just create a struct that holds all the components (and unsure that all have the same lifetime) and switch everything from Arc to just an immutable reference?

ultimaweapon:
> By the way, do all the components have to be in an Arc? Couldn't we just create a struct that holds all the components and switch everything from Arc to just an immutable reference?

The problem is we need static lifetime due to PS4 threading, which mean those objects must be either global or getting a static lifetime with `Box::leak`.

src/kernel/src/fs/mod.rs
```diff
@@ -193,13 +193,7 @@ impl Fs {
...
                     nd.cnd.nameptr += 1;
 
-                    if nd
-                        .cnd
-                        .pnbuf
-                        .get(nd.cnd.nameptr)
-                        .filter(|&v| *v == b'/')
-                        .is_none()
-                    {
+                    if nd.cnd.pnbuf.get(nd.cnd.nameptr).is_some_and(|&v| v == b'/') {
```
ultimaweapon, line 196: I don't think the refactored version has the same behavior.

src/kernel/src/fs/dev/console.rs
```diff
@@ -55,3 +67,17 @@ impl Display for Console {
         Self::PATH.fmt(f)
     }
 }
+
+#[derive(Debug, Error)]
+enum IoctlErr {
```
ultimaweapon, line 72: I never saw any error types in std having `Err` as a suffix.

src/kernel/src/process/session.rs
```diff
@@ -1,18 +1,43 @@
...
 /// An implementation of `session` structure.
 #[derive(Debug)]
 pub struct VSession {
-    id: NonZeroI32, // s_sid
-    login: String,  // s_login
+    id: NonZeroI32,               // s_sid
+    login: Gutex<String>,         // s_login
+    vnode: Option<Vnode>,         // s_ttyvp
+    tty: Gutex<Option<Arc<Tty>>>, // s_ttyp
+    gg: Arc<GutexGroup>,
```
ultimaweapon, line 12: We don't need to keep this if it is not using later on.

src/kernel/src/tty/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+        const TF_OPENED_IN = 0x00008;
+        const TF_OPENED_OUT = 0x00010;
+        const TF_OPENED_CONS = 0x00020;
+        const TF_OPENED = Self::TF_OPENED_IN.bits() | Self::TF_OPENED_OUT.bits() | Self::TF_OPENED_CONS.bits();
+        const TF_GONE = 0x00040;
+    }
+}
+
+#[derive(Debug, Error)]
+enum TtyErr {
```
ultimaweapon, line 113: Same as before.

VocalFan:
> > By the way, do all the components have to be in an Arc? Couldn't we just create a struct that holds all the components and switch everything from Arc to just an immutable reference?
> 
> The problem is we need static lifetime due to PS4 threading, which mean those objects must be either global or getting a static lifetime with `Box::leak`.

I just helped make a PR for SlimeVR... Their method to keep something is to use leak.

However, they do it a different way. `std::mem::forget(job)`

src/kernel/src/fs/vnode.rs
```diff
@@ -53,6 +57,7 @@ impl Drop for Vnode {
 #[derive(Debug)]
 pub enum VnodeType {
     Directory(bool),
+    Bad,
```
ultimaweapon, line 60: Making something in an invalid state is not work well with Rust. So let's remove this.

ultimaweapon (changes requested):

ultimaweapon:
> > > By the way, do all the components have to be in an Arc? Couldn't we just create a struct that holds all the components and switch everything from Arc to just an immutable reference?
> > 
> > 
> > The problem is we need static lifetime due to PS4 threading, which mean those objects must be either global or getting a static lifetime with `Box::leak`.
> 
> I just helped make a PR for SlimeVR... Their method to keep something is to use leak.
> 
> However, they do it a different way. `std::mem::forget(job)`

We have tried `Box::leak` before we switched to `Arc`. `forget` for a different purpose.

src/kernel/src/fs/vnode.rs
SuchAFuriousDeath: So... how do we represent this state? Option<Vnode>?

src/kernel/src/fs/dev/console.rs
SuchAFuriousDeath: I'ff fix it.

src/kernel/src/fs/mod.rs
SuchAFuriousDeath: Yeah, you're right.

src/kernel/src/process/session.rs
SuchAFuriousDeath: I didn't realize we don't have to hold on to it, because it's an Arc :)

src/kernel/src/fs/vnode.rs
ultimaweapon: The vnode itself should never be in that state for us. That mean it does not exists.

src/kernel/src/tty/mod.rs
SuchAFuriousDeath: Fixed

src/kernel/src/process/session.rs
SuchAFuriousDeath: Fixed

src/kernel/src/fs/vnode.rs
SuchAFuriousDeath: Oh

src/kernel/src/fs/dev/console.rs
SuchAFuriousDeath: Fixed

src/kernel/src/fs/mod.rs
SuchAFuriousDeath: Hopefully fixed

src/kernel/src/tty/mod.rs
ultimaweapon: It still has `Err` as a suffix.

SuchAFuriousDeath: That's weird, I remember renaming it. Anyway, not it should be fixed for real :D

src/kernel/src/process/session.rs
```diff
@@ -1,18 +1,41 @@
...
+use std::{num::NonZeroI32, sync::Arc};
 
 /// An implementation of `session` structure.
 #[derive(Debug)]
 pub struct VSession {
-    id: NonZeroI32, // s_sid
-    login: String,  // s_login
+    id: NonZeroI32,               // s_sid
+    login: Gutex<String>,         // s_login
+    vnode: Option<Vnode>,         // s_ttyvp
```
ultimaweapon, line 10: Do you know why FreeBSD keep this?

ultimaweapon (changes requested):

src/kernel/src/process/session.rs
SuchAFuriousDeath: I don't

src/kernel/src/fs/dev/mod.rs
```diff
@@ -70,6 +70,34 @@ pub fn make_dev<N: Into<String>>(
     Ok(dev)
 }
 
+/// See `prep_devname` on the PS4 for a reference.
+fn prepare_name(name: String) -> Result<String, MakeDevError> {
+    if name.len() > 63 {
+        return Err(MakeDevError::NameTooLong);
+    }
+
+    let name = name.trim_start_matches('/').to_string();
```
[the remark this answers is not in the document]
SuchAFuriousDeath: I think it is, but it will take some time

src/kernel/src/process/session.rs
ultimaweapon: How this field is set?

SuchAFuriousDeath:
> > > > By the way, do all the components have to be in an Arc? Couldn't we just create a struct that holds all the components and switch everything from Arc to just an immutable reference?
> > > 
> > > 
> > > The problem is we need static lifetime due to PS4 threading, which mean those objects must be either global or getting a static lifetime with `Box::leak`.
> > 
> > 
> > I just helped make a PR for SlimeVR... Their method to keep something is to use leak.
> > However, they do it a different way. `std::mem::forget(job)`
> 
> We have tried `Box::leak` before we switched to `Arc`. `forget` for a different purpose.

I google this and maybe we could use scoped threads?
 https://doc.rust-lang.org/stable/std/thread/fn.scope.html
 
 'Unlike non-scoped threads, scoped threads can borrow non-'static data, as the scope guarantees all threads will be joined at the end of the scope.'

ultimaweapon:
Scoped-thread cannot be used to implement threading syscall.

SuchAFuriousDeath:
Why is that, if I may ask?

ultimaweapon:
Let's say we implement [thr_new](https://man.freebsd.org/cgi/man.cgi?query=thr_new&sektion=2&n=1). How do we return to the caller without killing the spawned thread within the scope?

SuchAFuriousDeath:
Oh, right. I misunderstood what you meant by 'threading syscall'.

SuchAFuriousDeath:
But somehow, intuitevely, it feels like it should be possible to do, somehow.

SuchAFuriousDeath:
If we could somehow hold onto the scope as a field of VProc or something like that, the thread wouldn't die, but it would be guaranteed not to outlive the instance of VProc I don't know if something like that is possible though

ultimaweapon:
The point of scoped thread is to prevent any threads inside to leak. That mean it is not possible.

SuchAFuriousDeath:
I thought about this more and it seems you're right :((((

SuchAFuriousDeath:
By the way, don't you think it would be better to split our handle_64 function into handle_i64 and handle_u64 so that we don't have to change types?

ultimaweapon:
We can handled both types in `handle_64`.

src/kernel/src/fs/dev/mod.rs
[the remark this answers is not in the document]
SuchAFuriousDeath: This isn't actually so easy, at least to me. How would you do it?

src/kernel/src/process/session.rs
SuchAFuriousDeath: "Vnode of controlling tty".

SuchAFuriousDeath: http://fxr.watson.org/fxr/source/fs/devfs/devfs_vnops.c?v=FREEBSD-9-STABLE#L776

src/kernel/src/fs/dev/mod.rs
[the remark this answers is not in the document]
ultimaweapon: Then no need to do this for now.

src/kernel/src/process/session.rs
ultimaweapon: In that case better to merge those fields together so it will be behind a single `Option`.

src/kernel/src/fs/dev/mod.rs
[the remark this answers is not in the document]
SuchAFuriousDeath: Aren't all the device names known at compile time? If so, we could do this with a macro.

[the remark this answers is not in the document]
ultimaweapon: Good idea. But I think it will be too big for this PR (actually it is already big).

src/kernel/src/process/session.rs
SuchAFuriousDeath: You mean s_ttydp and s_ttyvp? A s_ttydp equivalent isn't in our code currently.

ultimaweapon: I mean `vnode` and `tty` because it seems like both of them must be `Some` or `None` at a given time.

SuchAFuriousDeath: image
Is this ok? I don't know how else to do this. Having it like this makes it map very weirdly to the FreeBSD code.

ultimaweapon: Sometime it will look weird to FreeBSD code but it is much more fitted with Rust so don't worry about that. In the end we will not care if our code similar to FreeBSD or not as long as it archive the same behavior.

SuchAFuriousDeath: The current code now panics though :D

SuchAFuriousDeath: Actually. I don' really think the fields should be together. in one option. 
image
The first line triggers the tty ioctl and only after that, the vnode is set on the second-to-last line. But the tty ioctl could return an error and the vnode won't be set in that case.

ultimaweapon: You mean the TTY will be set alone in the d_ioctl?

SuchAFuriousDeath: Actually, no. I'm just dumb. I think I understand what needs to be done. But even with that, I'll have to wait for the fs refactor to finalize, because I don't have access to the vnode right now because we're not there yet with the fs rewrite

ultimaweapon: Probably use `Fs::root` for a vnode for now then replace it later.

SuchAFuriousDeath: So I have to pass Arc<Fs> all the way down to tty? :(

ultimaweapon: Okay I see what you mean. In this case just remove vnode from here should be okay.

ultimaweapon:
I got a bad news (or good news?). My next PR is going to be overthrow the current FS structure by start migrating to the new FS. That mean the easier way for this PR to migrate to a new FS is by start a clean branch and copy part by part from the old branch.

SuchAFuriousDeath:
Nice (and not nice at the same time :D).

PR #536, opened 2023-12-29
Fixes incorrect credential for process

ultimaweapon:


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
We could create a wrapper, something like 'NonEmptyVec' to be able to get rid of the assert.

PR #554, opened 2024-01-13
Initializes nullfs implementation

SuchAFuriousDeath:


src/kernel/src/fs/vnode.rs
```diff
@@ -138,19 +146,22 @@ pub struct VopVector {
     pub default: Option<&'static Self>, // vop_default
     pub access: Option<VopAccess>,      // vop_access
     pub accessx: Option<VopAccessX>,    // vop_accessx
+    pub bypass: Option<VopBypass>,      // vop_bypass
```
ultimaweapon, line 149: Have you found a solution for this as I stated in the struct docs?

ultimaweapon (changes requested):

src/kernel/src/fs/vnode.rs
SuchAFuriousDeath: Uh, I completely missed that :sweat_smile:

SuchAFuriousDeath: So, just remove it?

ultimaweapon: Yeah unless you have a solution for this.

SuchAFuriousDeath: Removed.

src/kernel/src/fs/mount.rs
```diff
@@ -138,15 +143,163 @@ bitflags! {
     pub struct MountFlags: u64 {
         const MNT_RDONLY = 0x0000000000000001;
         const MNT_NOSUID = 0x0000000000000008;
+        const MNTK_LOOKUP_EXCL_DOTDOT = 0x0000000000000800;
```
ultimaweapon, line 146: Where is this defined in the FreeBSD?

src/kernel/src/fs/mount.rs
```diff
@@ -138,15 +143,163 @@ bitflags! {
...
+    pub fn retain(&mut self, mut f: impl FnMut(&&'static str, &mut MountOpt) -> bool) {
+        self.0.retain(|k, v| f(k, v));
+    }
+}
+
+pub(super) enum MountOpt {
+    Bool(bool),
+    Int(i32),
+    Str(Box<str>),
+    VPathBuf(VPathBuf),
```
ultimaweapon, line 178: `VPath` probably a better name here.

src/kernel/src/fs/mount.rs
```diff
@@ -138,15 +143,163 @@ bitflags! {
...
+    PathBuf(PathBuf),
```
ultimaweapon, line 179: `Path` probably a better name here.

src/kernel/src/fs/mount.rs
```diff
@@ -138,15 +143,163 @@ bitflags! {
...
+
+    pub fn insert(&mut self, k: &'static str, v: impl Into<MountOpt>) {
+        self.0.insert(k.into(), v.into());
+    }
+
+    pub fn remove(&mut self, k: &'static str) -> Option<MountOpt> {
+        self.0.remove(k)
+    }
+
+    pub fn retain(&mut self, mut f: impl FnMut(&&'static str, &mut MountOpt) -> bool) {
```
ultimaweapon, line 169: Why do we need `&&'static str`?

ultimaweapon (changes requested):

src/kernel/src/fs/mount.rs
SuchAFuriousDeath: http://fxr.watson.org/fxr/source/sys/mount.h?v=FREEBSD-9-STABLE#L376

src/kernel/src/fs/mount.rs
SuchAFuriousDeath: Fixed

src/kernel/src/fs/mount.rs
SuchAFuriousDeath: Done

src/kernel/src/fs/mount.rs
SuchAFuriousDeath: Done

src/kernel/src/fs/mount.rs
ultimaweapon: Our `MountFlags` here represents a flags in `mnt_flag`, not `mnt_kern_flag`.

SuchAFuriousDeath: Oops. removed.

SuchAFuriousDeath:
Small update: I discovered the mysterious ioctl command in sys_ioctl. It's the ones in this file: http://fxr.watson.org/fxr/source/sys/filio.h?v=FREEBSD-9-STABLE.

SuchAFuriousDeath:
@ultimaweapon This is the current error by the way:
image

src/kernel/src/tty/mod.rs
```diff
@@ -16,7 +16,7 @@ pub struct TtyManager {
 impl TtyManager {
     const TIOCSCTTY: IoCmd = IoCmd::io(b't', 97);
 
-    pub fn new(fs: &Arc<Fs>) -> Result<Arc<Self>, TtyError> {
+    pub fn new(_fs: &Arc<Fs>) -> Result<Arc<Self>, TtyError> {
```
ultimaweapon, line 19: Better to remove the whole argument since it does not used. Not sure why I added this argument.

src/kernel/src/fs/vnode.rs
```diff
@@ -128,10 +129,13 @@ impl Drop for Vnode {
 }
 
 /// An implementation of `vtype`.
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
```
ultimaweapon, line 132: `VnodeType` should not implement `Copy` because we embed additional data in the variants.

src/kernel/src/fs/perm.rs
```diff
@@ -137,7 +137,8 @@ impl From<Mode> for u32 {
 bitflags! {
     /// An implementation of `accmode_t`.
     pub struct Access: u32 {
-        const EXEC = 000000000100; // VEXEC
+        const EXEC  = 000000000100; // VEXEC
+        const WRITE = 000000000200; // VWRITE
```
ultimaweapon, line 141: Decimal?

src/kernel/src/fs/tmp/mod.rs
```diff
@@ -1,14 +1,13 @@
...
 use crate::fs::Mode;
 use crate::ucred::{Gid, Uid};
-use std::any::Any;
-use std::collections::HashMap;
 use std::num::NonZeroI32;
 use std::sync::Arc;
 use thiserror::Error;
 
-fn mount(mnt: &mut Mount, mut opts: HashMap<String, Box<dyn Any>>) -> Result<(), Box<dyn Errno>> {
+#[allow(unused_variables)]
```
ultimaweapon, line 9: This one should not applied here because it is a WIP function.

src/kernel/src/fs/mount.rs
```diff
@@ -153,8 +154,227 @@ bitflags! {
...
+    fn from(v: Uid) -> Self {
+        Self::Uid(v)
+    }
+}
+
+impl From<Mode> for MountOpt {
+    fn from(v: Mode) -> Self {
+        Self::Mode(v)
+    }
+}
```
ultimaweapon, line 263: These should not be here otherwise this file will be polluted with all of types in the kernel.

src/kernel/src/fs/mount.rs
```diff
@@ -153,8 +154,227 @@ bitflags! {
...
+    fn from(v: String) -> Self {
+        Self::Str(v.into_boxed_str())
+    }
+}
+
+impl From<VPathBuf> for MountOpt {
+    fn from(v: VPathBuf) -> Self {
+        Self::VPath(v)
+    }
+}
```
ultimaweapon, line 233: Same here.

src/kernel/src/fs/mount.rs
```diff
@@ -153,8 +154,227 @@ bitflags! {
...
+impl TryFrom<MountOpt> for VPathBuf {
+    type Error = ();
+
+    fn try_from(v: MountOpt) -> Result<Self, Self::Error> {
+        match v {
+            MountOpt::VPath(v) => Ok(v),
+            _ => Err(()),
+        }
+    }
+}
```
ultimaweapon, line 307: Same here.

src/kernel/src/fs/mount.rs
```diff
@@ -153,8 +154,227 @@ bitflags! {
...
+impl TryFrom<MountOpt> for Mode {
+    type Error = ();
+
+    fn try_from(v: MountOpt) -> Result<Self, Self::Error> {
+        match v {
+            MountOpt::Mode(v) => Ok(v),
+            _ => Err(()),
+        }
+    }
+}
```
ultimaweapon, line 362: Same here.

src/kernel/src/fs/mod.rs
```diff
@@ -147,14 +144,20 @@ impl Fs {
...
     }
 
-    pub fn open<P: AsRef<VPath>>(&self, path: P, td: Option<&VThread>) -> Result<VFile, OpenError> {
-        todo!()
+    pub fn open(
+        &self,
+        _path: impl AsRef<VPath>,
+        _td: Option<&VThread>,
+    ) -> Result<VFile, OpenError> {
+        let _vnode = self.lookup(_path, _td).map_err(OpenError::LookupFailed)?;
```
ultimaweapon, line 152: `_path` and `_td` is already used so no reasons to prefixed it with `_`.

ultimaweapon (changes requested):

src/kernel/src/fs/mount.rs
SuchAFuriousDeath: So... you want me to move them to where every type is defined?

SuchAFuriousDeath:
@ultimaweapon The enum_delegate crate can derive all the boilerplate conversion for us.

ultimaweapon:
> @ultimaweapon The enum_delegate crate can derive all the boilerplate conversions for us. (I just discovered it)

Is there other crate just for conversion? `enum_delegate` has additional features that its syntax quite cryptic, which we should avoid using it. If we introduced this crate here people might accidentally using it.

SuchAFuriousDeath:
I didn't find another one, maybe there is one.

SuchAFuriousDeath:
Maybe this one: https://lib.rs/crates/enum-conversion-derive. I'll try it out.

ultimaweapon:
> Maybe this one: https://lib.rs/crates/enum-conversion-derive. I'll try it out.

This one is GPL, which is a problem for us. If there is no other crate available we probably need to write our own version.

SuchAFuriousDeath:
I managed to implement it myself.

SuchAFuriousDeath:
What's amazing is that we can return the opt back as the Err variant, allowing us to reuse it for error handling. I reworked the remove function to take advantage of this, producing much better error values.

src/kernel/Cargo.toml
```diff
@@ -14,6 +14,7 @@ byteorder = "1.4"
 clap = { version = "4.1", features = ["derive"] }
 discord-rich-presence = "0.2"
 elf = { path = "../elf" }
+enum_delegate = "0.2.0"
```
ultimaweapon, line 17: Why we still need this?

src/kernel/src/fs/perm.rs
```diff
@@ -137,7 +137,8 @@ impl From<Mode> for u32 {
 bitflags! {
     /// An implementation of `accmode_t`.
     pub struct Access: u32 {
-        const EXEC = 000000000100; // VEXEC
+        const EXEC  = 0x000000000100; // VEXEC
+        const WRITE = 0x000000000200; // VWRITE
```
ultimaweapon, line 141: Actually it is octal, not hexadecimal.

src/macros/src/lib.rs
```diff
@@ -14,6 +15,15 @@ pub fn cpu_abi(_: TokenStream, item: TokenStream) -> TokenStream {
         .into()
 }
 
+#[proc_macro_attribute]
```
ultimaweapon, line 18: Do we really need an attribute macro? I think just a derive macro should be enough.

ultimaweapon (changes requested):

src/kernel/src/fs/mount.rs
```diff
@@ -140,8 +150,94 @@ bitflags! {
...
+
+#[derive(Debug, Error)]
+pub enum MountOptError {
+    #[error("mount opt \"{optname}\" is of wrong type: expected {expected}, got: {got:?}")]
+    UnexpectedOptType {
+        optname: &'static str,
+        expected: &'static str,
+        got: MountOpt,
+    },
```
ultimaweapon, line 225: If `MountOptError` never going to have more than one variant better to change it to a struct and implement `Error` manually.

src/kernel/src/fs/dev/mod.rs
```diff
@@ -326,7 +324,7 @@ pub fn mount(
     cred: &Arc<Ucred>,
     path: VPathBuf,
     parent: Option<Arc<Vnode>>,
-    opts: HashMap<String, Box<dyn Any>>,
+    _opts: MountOpts,
```
ultimaweapon, line 327: Better to use `_` here because `devfs` does not support options. Not sure why I keep it here.

ultimaweapon (changes requested):

src/kernel/src/fs/null/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+fn root(_mnt: &Arc<Mount>) -> Arc<Vnode> {
+    todo!()
+}
+
+pub(super) static NULLFS_OPS: FsOps = FsOps { root };
+
+#[derive(Debug, Error)]
+#[allow(dead_code)]
```
ultimaweapon, line 54: This one should be only applied to the PS4 types that we need to keep any discovered fields.

src/kernel/src/fs/null/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        .remove("target")
+        .or_else(|| opts.remove("from"))
+        .ok_or(MountError::NoTarget)?
+        .unwrap();
+
+    if target.is_empty() {
+        Err(MountError::EmptyTarget)?;
+    }
+
+    let _target: &VPath = target.as_ref().try_into().unwrap();
```
ultimaweapon, line 42: If we accept `target` and `from` as a `VPathBuf` we can eliminate this. We also don't need the above statement that check if target is empty.

src/kernel/src/fs/null/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+    RootFs,
+
+    #[error("update mount is not supported without export option")]
+    NoExport,
+
+    #[error("target path is not specified")]
+    NoTarget,
+
+    #[error("target path is empty")]
+    EmptyTarget,
```
ultimaweapon, line 66: We don't need this if we followed my previous idea.

src/kernel/src/fs/null/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// An implementation of `null_mount` structure.
+#[allow(dead_code)]
+struct NullFs {
+    root: Arc<Vnode>,          // nullm_rootvp
+    lower: Option<Arc<Vnode>>, // nullm_lowervp
+    flags: NullFsFlags,        // null_flags
+}
+
+#[allow(dead_code)]
```
ultimaweapon, line 96: Same as before.

ultimaweapon (changes requested):

src/kernel/src/fs/null/vnode.rs
```diff
@@ -0,0 +1,80 @@
+use super::NullFs;
+use crate::{
+    errno::{Errno, EROFS},
+    fs::{perm::Access, MountFlags, OpenFlags, VFile, Vnode, VnodeType, VopVector},
+    process::VThread,
+};
+use std::{num::NonZeroI32, sync::Arc};
+use thiserror::Error;
+
+#[allow(dead_code)]
```
ultimaweapon, line 10: Why do we need this?

src/kernel/src/fs/null/vnode.rs
```diff
@@ -0,0 +1,80 @@
...
+pub(super) static VNODE_OPS: VopVector = VopVector {
+    default: None,
+    access: Some(access),
+    accessx: Some(access),
+    getattr: None,
```
ultimaweapon, line 15: This one is not null on the PS4 so better to add it with `todo` instead of leaving this.

src/kernel/src/fs/null/vnode.rs
```diff
@@ -0,0 +1,80 @@
...
+pub enum BypassError {}
+
+impl Errno for BypassError {
+    fn errno(&self) -> NonZeroI32 {
+        todo!()
+    }
+}
+
+fn lookup(vn: &Arc<Vnode>, td: Option<&VThread>, name: &str) -> Result<Arc<Vnode>, Box<dyn Errno>> {
+    let null_mount: &NullFs = vn.data().downcast_ref().unwrap();
```
ultimaweapon, line 60: This does not seems right. Usually what stored here is not a mount data but a file or directory data.

ultimaweapon (changes requested):

SuchAFuriousDeath:
Are you waiting for me to merge the backend PR?

src/kernel/src/fs/null/vnode.rs
```diff
@@ -33,6 +32,36 @@ fn access(vn: &Arc<Vnode>, _td: Option<&VThread>, access: Access) -> Result<(),
     todo!();
 }
 
+fn getattr(vn: &Arc<Vnode>) -> Result<VnodeAttrs, Box<dyn Errno>> {
+    todo!()
+}
+
+fn lookup(vn: &Arc<Vnode>, td: Option<&VThread>, name: &str) -> Result<Arc<Vnode>, Box<dyn Errno>> {
+    let null_mount: &NullNode = vn.data().downcast_ref().unwrap();
```
ultimaweapon, line 40: `null_mount`?

src/kernel/src/fs/null/vnode.rs
```diff
@@ -33,6 +32,36 @@ fn access(vn: &Arc<Vnode>, _td: Option<&VThread>, access: Access) -> Result<(),
...
+
+    let lower = null_mount
```
ultimaweapon, line 42: There is a PS4 specific code before this one so better to put a `TODO` here.

src/kernel/src/fs/null/vnode.rs
```diff
@@ -33,6 +32,36 @@ fn access(vn: &Arc<Vnode>, _td: Option<&VThread>, access: Access) -> Result<(),
...
+        .lower()
+        .lookup(td, name)
+        .map_err(LookupError::LookupFailed)?;
+
+    let vnode = if Arc::ptr_eq(&lower, vn) {
```
ultimaweapon, line 47: Is this even possible? The `vn` here was produced by nullfs and the lower one we got here should always coming from elsewhere.

src/kernel/src/fs/null/vnode.rs
```diff
@@ -56,25 +85,14 @@ impl Errno for BypassError {
...
-        vn.clone()
-    } else {
-        todo!();
-    };
-
-    Ok(vnode)
+#[derive(Debug, Error)]
+pub enum LookupError {
+    #[error("lookup failed")]
+    LookupFailed(#[source] Box<dyn Errno>),
```
ultimaweapon, line 91: I think `LookupFromLowerFailed` are more descriptive.

ultimaweapon (changes requested):

src/kernel/src/fs/null/vnode.rs
SuchAFuriousDeath: oops

src/kernel/src/fs/null/vnode.rs
SuchAFuriousDeath: I already asked about it I think in one of your PR's we don't support cn_nameiop. That code only has to do with that if I'm not mistaken.

src/kernel/src/fs/null/vnode.rs
SuchAFuriousDeath: Not sure

SuchAFuriousDeath:
WIll we implement VOP_ISLOCKED and VOP_UNLOCK?

src/kernel/src/fs/null/vnode.rs
ultimaweapon: Sorry I mistaken this one is `open`.

src/kernel/src/fs/null/vnode.rs
ultimaweapon: @SuchAFuriousDeath I mean the variant, not enum itself.

ultimaweapon:
> WIll we implement VOP_ISLOCKED and VOP_UNLOCK or will it be just be a no-op for us?

I did not take a look what it does yet. Do you know how it is working?

src/kernel/src/fs/null/vnode.rs
SuchAFuriousDeath: oh :D

SuchAFuriousDeath:
Nope.

SuchAFuriousDeath:
@ultimaweapon I think we should just make a figure out what null_bypass would do manually and implement every instance as a separate function. 
As a sidenote, why does make_dev take an &Arc<Cdevsw>? I was under the impression that all these devices were statically in memory.

SuchAFuriousDeath:
If we allow the mount function to return an Arc<Mount>, then I think our problem is solved.

SuchAFuriousDeath:
As another sidenote, I don't think Vattrs:new is a good idea long-terim. I think we should: 
1) make all the fields public and do something like
Vattrs {fsid, ...Vattrs::null() }
2) or alternatively
Vattrs::null().with_fsid();

I think specifying every field will be an extreme headache

SuchAFuriousDeath:
@ultimaweapon Another sidenote: I think the lookup functions should take a componentname or something like that. i think it's easy to assume that name: &str in a function called 'lookup' is the full path if you don't know the context. Also, won't we need cn_nameiop in the future to support renaming and deleting (and IS_LAST_CN)? Even in null_lookup, there are checks for this. 
Of course I read the doc comment which mentions that we don't implement nameidata because it's too compilcated, but to me it's not obvious whether we completely discard even componentname. Let me know your stance on this.
That being said, I'm not fully familiar with how nameidata and componentname work, so maybe I'm missing something?
Rust has this: https://doc.rust-lang.org/stable/std/path/struct.Components.html Where the component is https://doc.rust-lang.org/stable/std/path/enum.Component.html. Maybe we could do something similar with VPath::components.

SuchAFuriousDeath:
> @ultimaweapon Another sidenote: I think the lookup functions should take a componentname or something like that. i think it's easy to assume that name: &str in a function called 'lookup' is the full path if you don't know the context (we could rename the argument, but I think it's better if the type itself was name Component or VComponent, because in order to use it, you'd have to look up the definition and methods). Also, won't we need cn_nameiop in the future to support renaming and deleting (and IS_LAST_CN)? Even in null_lookup, there are checks for this. We already talked about this. Of course I read the doc comment which mentions that we don't implement nameidata because it's too compilcated, but to me it's not obvious whether we completely discard even componentname. Let me know your stance on this. I'm not fully familiar with how nameidata and componentname work, so maybe I'm missing something? Rust has this: https://doc.rust-lang.org/stable/std/path/struct.Components.html Where the component is https://doc.rust-lang.org/stable/std/path/enum.Component.html. Maybe we could do something similar with VPath::components.

I'm willing to implement this if you're up for it.

ultimaweapon:
> > @ultimaweapon Another sidenote: I think the lookup functions should take a componentname or something like that. i think it's easy to assume that name: &str in a function called 'lookup' is the full path if you don't know the context (we could rename the argument, but I think it's better if the type itself was name Component or VComponent, because in order to use it, you'd have to look up the definition and methods). Also, won't we need cn_nameiop in the future to support renaming and deleting (and IS_LAST_CN)? Even in null_lookup, there are checks for this. We already talked about this. Of course I read the doc comment which mentions that we don't implement nameidata because it's too compilcated, but to me it's not obvious whether we completely discard even componentname. Let me know your stance on this. I'm not fully familiar with how nameidata and componentname work, so maybe I'm missing something? Rust has this: https://doc.rust-lang.org/stable/std/path/struct.Components.html Where the component is https://doc.rust-lang.org/stable/std/path/enum.Component.html. Maybe we could do something similar with VPath::components.
> 
> I'm willing to implement this if you're up for it. Of course, we don't have to implement 100% of what FreeBSD does with it.

I don't have any problems if you can implement a better solution. But this PR is already too big so you should not put more things on this. Big PR is not good for everyone because it will be harder to get merge and it also have more chance to block other PRs.

SuchAFuriousDeath:
Of course, I meant in another PR.

SuchAFuriousDeath:
I think Mount::with_data_fn is actually unsound and Arc::get_mut will fail :tired_face: I don't really know how to construct this :(

src/kernel/src/fs/perm.rs
```diff
@@ -136,8 +136,26 @@ impl From<Mode> for u32 {
...
+        const READ_ATTRIBUTES  = 0o00002000000; // VREAD_ATTRIBUTES
+        const WRITE_ATTRIBUTES = 0o00004000000; // VWRITE_ATTRIBUTES
+        const DELETE           = 0o00010000000; // VDELETE_CHILD
+        const READ_ACL         = 0o00020000000; // VREAD_ACL
+        const WRITE_ACL        = 0o00040000000; // VWRITE_ACL
+        const WRITE_OWNER      = 0o00100000000; // VWRITE_OWNER
+        const SYNCHRONIZE      = 0o00200000000; // VSYNCHRONIZE
+
+        const ADMIN_PERMS = Self::ADMIN.bits() | Self::WRITE_ATTRIBUTES.bits() | Self::WRITE_ACL.bits() | Self::WRITE_OWNER.bits();
+        const STAT_PERMS = Self::READ_ATTRIBUTES.bits() | Self::READ_ACL.bits();
```
ultimaweapon, line 158: Have you verified these with the PS4? If not better to remove it since the PS4 might be different. We already have this issue with `mmap` before that the value is not the same as FreeBSD.

ultimaweapon (changes requested):

src/kernel/src/fs/perm.rs
SuchAFuriousDeath: Most of these are used in unixify_access and I doubt that. I can remove the rest if you want

ultimaweapon: You can use `unixify_access` on the PS4 to verified if those values are actually the same. Then remove all of non-verified ones.

src/kernel/src/ee/native/mod.rs
```diff
@@ -302,9 +302,11 @@ impl NativeEngine {
                 target[4] = tp[3];
 
                 // Path the remaining with "int3".
-                for i in 5..inst.len() {
-                    target[i] = 0xcc;
-                }
+                target
+                    .iter_mut()
+                    .take(inst.len())
+                    .skip(5)
```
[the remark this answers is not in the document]
SuchAFuriousDeath: It was suggested by clippy :flushed:

src/kernel/src/fs/perm.rs
SuchAFuriousDeath: If this:
image
is equivalent to access.intersects(Access::EXPLICIT_DENY), then it is correct. I removed the two unused constants

src/kernel/src/arnd.rs
```diff
@@ -36,6 +33,12 @@ impl Arnd {
         s.sbox.swap(s.i as usize, s.j as usize);
         s.sbox[s.sbox[s.i as usize].wrapping_add(s.sbox[s.j as usize]) as usize]
     }
+
+    /// An implementation of `arc4random` on the PS4.
+    /// TODO: Implement reseed.
+    pub fn rand_bytes(buf: &mut [u8]) {
```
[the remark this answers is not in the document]
SuchAFuriousDeath: Hopefully I understood what you meant. You can take a look.

src/kernel/src/ee/native/mod.rs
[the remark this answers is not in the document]
ultimaweapon: Seems like a clippy bug here.

src/kernel/src/arnd.rs
```diff
@@ -1,28 +1,31 @@
...
+/// TODO: Implement reseed.
+pub fn rand_bytes(buf: &mut [u8]) {
+    ARND.rand_bytes_internal(buf)
+}
 
 /// Random number generator based on
 /// https://github.com/freebsd/freebsd-src/blob/release/9.1.0/sys/libkern/arc4random.c.
 #[derive(Debug)]
-pub struct Arnd {
+struct Arnd {
```
ultimaweapon, line 13: Now we don't need this. Better to remove it and work on `State` directly.

src/kernel/src/main.rs
```diff
@@ -33,6 +32,8 @@ use sysinfo::{MemoryRefreshKind, System};
 use thiserror::Error;
 use tty::TtyInitError;
 
+pub use self::arnd::rand_bytes;
+
```
ultimaweapon, line 36: I don't think we should re-export any sub-module items at the top-level application.

src/kernel/src/ee/native/mod.rs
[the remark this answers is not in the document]
ultimaweapon: Wait, seems like clippy is correct here.

SuchAFuriousDeath:
I got too tangled up in this. We don't even know if we actually need the data to hold a reference to the mount.
Now it should be semi-usable.
It would probably to create a trait-based implementation of VfsOps as well, I absolutely love VnodeBacked, it was a brilliant idea.

src/kernel/src/arnd.rs
```diff
@@ -1,47 +1,60 @@
...
+/// TODO: Implement reseed.
+pub fn rand_bytes(buf: &mut [u8]) {
+    ARND.lock().unwrap().rand_bytes(buf)
+}
 
 /// Random number generator based on
 /// https://github.com/freebsd/freebsd-src/blob/release/9.1.0/sys/libkern/arc4random.c.
 #[derive(Debug)]
-pub struct Arnd {
+struct Arnd {
```
ultimaweapon, line 12: Why do we still need this?

src/kernel/src/fs/vnode.rs
```diff
@@ -208,6 +222,12 @@ impl VnodeAttrs {
     pub fn mode(&self) -> Mode {
         self.mode
     }
+
+    pub fn with_fsid(mut self, fsid: u32) -> Self {
+        self.fsid = fsid;
+
+        self
+    }
```
ultimaweapon, line 230: Why we use this instead of `set_fsid(&mut self)`?

src/kernel/src/fs/dev/vnode.rs
```diff
@@ -88,10 +88,11 @@ impl crate::fs::VnodeBackend for VnodeBackend {
         let mode = dirent.mode();
         let size = match vn.ty() {
             VnodeType::Directory(_) => 512,
-            VnodeType::Character => 0,
+            VnodeType::Link => todo!(), /* TODO: strlen(dirent.de_symlink) */
+            _ => 0,
         };
 
-        Ok(VnodeAttrs::new(*uid, *gid, *mode, size))
+        Ok(VnodeAttrs::new(*uid, *gid, *mode, size, 0))
```
ultimaweapon, line 95: Does the PS4 set fsid to zero here or we just use zero to represents not yet implemented? If it is the latter case better to replace the whole line with `todo` macro instead.

src/kernel/src/fs/file.rs
```diff
@@ -8,13 +8,15 @@ use std::sync::Arc;
...
 #[derive(Debug)]
+#[allow(dead_code)]
 pub struct VFile {
     ty: VFileType,                    // f_type
     data: Arc<dyn Any + Send + Sync>, // f_data
     ops: &'static VFileOps,           // f_ops
     flags: VFileFlags,                // f_flag
 }
 
+#[allow(dead_code)]
```
ultimaweapon, line 19: This should not apply on the implementation because it can hide the things that we don't actually need.

src/kernel/src/fs/host/vnode.rs
```diff
@@ -34,18 +34,16 @@ impl crate::fs::VnodeBackend for VnodeBackend {
...
         // TODO: Check how the PS4 assign file permissions for exfatfs.
         let mode = match vn.ty() {
             VnodeType::Directory(_) => Mode::new(0o555).unwrap(),
-            VnodeType::Character => unreachable!(), // The character device should only be in the devfs.
+            VnodeType::File | VnodeType::Link => todo!(),
+            VnodeType::Character => unreachable!(), // Character devices should only be in devfs.
         };
 
-        Ok(VnodeAttrs::new(Uid::ROOT, Gid::ROOT, mode, size))
+        Ok(VnodeAttrs::new(Uid::ROOT, Gid::ROOT, mode, size, 0))
```
ultimaweapon, line 46: Same as the previous question.

ultimaweapon (changes requested):

src/kernel/src/arnd.rs
SuchAFuriousDeath: I thought I removed it :eyes:

src/kernel/src/fs/host/vnode.rs
```diff
@@ -35,18 +35,16 @@ impl crate::fs::VnodeBackend for VnodeBackend {
...
         // TODO: Check how the PS4 assign file permissions for exfatfs.
         let mode = match vn.ty() {
             VnodeType::Directory(_) => Mode::new(0o555).unwrap(),
-            VnodeType::Character => unreachable!(), // The character device should only be in the devfs.
+            VnodeType::File | VnodeType::Link => todo!(),
+            VnodeType::Character => unreachable!(), // Character devices should only be in devfs.
         };
 
-        Ok(VnodeAttrs::new(Uid::ROOT, Gid::ROOT, mode, size))
+        Ok(VnodeAttrs::new(Uid::ROOT, Gid::ROOT, mode, size, u32::MAX))
```
ultimaweapon, line 47: Does `u32::MAX` as fsid are actually correct here?

src/kernel/src/fs/file.rs
ultimaweapon: Seems like to missed this one.

src/kernel/src/fs/host/vnode.rs
SuchAFuriousDeath: image

ultimaweapon: Seems like someone too lazy to implement it correctly here 😂

SuchAFuriousDeath: What do you mean? I thought we're doing what FreeBSD does.

SuchAFuriousDeath: To be honest, I have no idea what else I should do here

ultimaweapon: I mean someone at Sony are lazy to make implement this correctly. So we have no choice but follow it.

SuchAFuriousDeath: Oh. I thought that you meant me :sweat_smile:

src/kernel/src/fs/tmp/mod.rs
```diff
@@ -32,52 +29,37 @@ pub fn mount(
...
-        None => 0,
-    };
+    let size: usize = opts.remove("size").map_or(0, |v| v.unwrap());
 
     // Get maximum file size.
-    let file_size = match opts.remove("maxfilesize") {
-        Some(v) => *v.downcast::<u64>().unwrap(),
-        None => 0,
-    };
+    let file_size = opts.remove("file_size").map_or(0, |v| v.unwrap());
```
ultimaweapon, line 62: Wrong option name here.

src/kernel/src/fs/tmp/mod.rs
```diff
@@ -32,52 +29,37 @@ pub fn mount(
...
         attrs.uid()
     };
 
     // Get mode.
-    let mode = if cred.real_uid() == Uid::ROOT {
-        match opts.remove("mode") {
-            Some(v) => *v.downcast::<Mode>().unwrap(),
-            None => attrs.mode(),
-        }
+    let _mode = if cred.real_uid() == Uid::ROOT {
```
ultimaweapon, line 49: Better to not rename the variable name in a WIP function.

src/kernel/src/fs/tmp/mod.rs
```diff
@@ -32,52 +29,37 @@ pub fn mount(
...
         attrs.gid()
     };
 
     // Get UID.
-    let uid = if cred.real_uid() == Uid::ROOT {
-        match opts.remove("uid") {
-            Some(v) => *v.downcast::<Uid>().unwrap(),
-            None => attrs.uid(),
-        }
+    let _uid = if cred.real_uid() == Uid::ROOT {
```
ultimaweapon, line 42: Same here.

src/kernel/src/fs/tmp/mod.rs
```diff
@@ -32,52 +29,37 @@ pub fn mount(
     };
 
     // Get GID.
-    let gid = if cred.real_uid() == Uid::ROOT {
+    let _gid = if cred.real_uid() == Uid::ROOT {
```
ultimaweapon, line 32: Same here.

src/kernel/src/fs/null/vnode.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn errno(&self) -> NonZeroI32 {
+        match self {
+            Self::GetAttrFromLowerFailed(e) => e.errno(),
+        }
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum LookupError {
+    #[error("lookup failed")]
```
ultimaweapon, line 129: The message should be `lookup from lower vnode failed`

src/kernel/src/fs/mod.rs
```diff
@@ -660,6 +671,25 @@ impl Errno for LookupError {
     }
 }
 
+#[derive(Debug, Error)]
+#[allow(dead_code)]
```
ultimaweapon, line 675: Why do we need this?

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #557, opened 2024-01-14
Refresh memory info

### LINKED ISSUE #545: Memory Available become 0
```
Memory Available    : 0/0 MB
```

VocalFan:
Forces memory info to refresh/load. For some reason, it seems that sysinfo only initializes the value, but never grabs the information unless prompted.

Fixes #545

### DIFF
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -168,7 +168,11 @@ fn main() -> ExitCode {
         log,
         "Operating System    : {} {}",
         System::long_os_version().unwrap_or_else(|| "Unknown OS".to_string()),
-        System::kernel_version().unwrap_or_else(|| "Unknown Kernel".to_string())
+        if cfg!(target_os = "windows") {
+            System::kernel_version().unwrap_or_else(|| "Unknown Kernel".to_string())
+        } else {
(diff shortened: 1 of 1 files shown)


ultimaweapon:
Can you change it to use either `long_os_version` or `kernel_version`, not both? On Linux it just prints something like `Arch Linux Linux` when printing both, which is weird.

VocalFan:
> Can you change it to use either `long_os_version` or `kernel_version`, not both? On Linux it just prints something like `Arch Linux Linux` when printing both, which is weird.

Reason I have it like that is because on windows, this lets it print `Windows 11 Pro 22621`
Seeing as that is both the long_os_version (Windows 11 Pro), and Kernel Version (22621)

VocalFan:
Also for Linux, it's meant to be more like... `Linux Mint 21.3 5.15-LTS`

ultimaweapon:
`long_os_version` should be enough for us to diagnose the problem.

ultimaweapon:
Or split it into a separated line.

VocalFan:
How about it only gives the Kernel Version on Windows? ^

ultimaweapon (approved):

PR #560, opened 2024-01-14
Initial implementation of prison_priv_check and of sockets

SuchAFuriousDeath:


SuchAFuriousDeath:
Update: Initialized the implementation of sockets.

SuchAFuriousDeath:
@ultimaweapon I think falloc_budget will have to return an UninitVFile wrapper that can only be converted to VFile by providing the unitialized fields.

ultimaweapon:
> @ultimaweapon I think falloc_budget will have to return an UninitVFile wrapper that can only be converted to VFile by providing the unitialized fields or something like that.

Can we use the same pattern as FreeBSD here instead of introducing a new type?

SuchAFuriousDeath:
> > @ultimaweapon I think falloc_budget will have to return an UninitVFile wrapper that can only be converted to VFile by providing the unitialized fields or something like that.
> 
> Can we use the same pattern as FreeBSD here instead of introducing a new type?

I'm not sure, I'll revert some commits and leave that up to you. This PR already does a lot of things.

SuchAFuriousDeath:
Actually, looking at it now, we could probably merge falloc and finit into a single function, since finit can't fail.

SuchAFuriousDeath:
I just realized that I have a mistake here. I thought the vnode is just v_data, but there's actually a separate field for a vnode.

SuchAFuriousDeath:
@ultimaweapon This one is ready for a review now.

src/kernel/src/ee/llvm/codegen.rs
ultimaweapon, line 1: Better to remove the whole file since no one working on this part yet, which mean it is unclear what solution we will actually use.

src/kernel/src/ee/llvm/mod.rs
```diff
@@ -11,10 +11,12 @@ mod codegen;
 
 /// An implementation of [`ExecutionEngine`] using JIT powered by LLVM IR.
 #[derive(Debug)]
+#[allow(dead_code)]
```
ultimaweapon, line 14: Better to remove `llvm` field instead of applying this one because of the same previous reason.

src/kernel/src/ee/llvm/mod.rs
```diff
@@ -11,10 +11,12 @@ mod codegen;
...
 pub struct LlvmEngine {
     llvm: Arc<Llvm>,
 }
 
+#[allow(dead_code)]
```
ultimaweapon, line 19: Same here.

src/kernel/src/ee/llvm/mod.rs
```diff
@@ -53,6 +55,7 @@ impl LlvmEngine {
     }
 }
 
+#[allow(unused_variables)]
```
ultimaweapon, line 58: Same here.

src/kernel/src/ee/llvm/mod.rs
```diff
@@ -79,6 +82,7 @@ impl ExecutionEngine for LlvmEngine {
 #[derive(Debug)]
 pub struct RawFn {}
 
+#[allow(unused_variables)]
```
ultimaweapon, line 85: Same here.

src/kernel/src/llvm/module.rs
ultimaweapon, line 1: Better to remove this file too.

src/kernel/src/ucred/mod.rs
```diff
@@ -1,25 +1,40 @@
...
 use crate::errno::{Errno, EPERM};
+use crate::net::AddressFamily;
+use crate::ucred::prison::PrisonAllow;
+use crate::ucred::prison::PrisonFlags;
 use std::num::NonZeroI32;
 use thiserror::Error;
 
 mod auth;
 mod id;
+pub mod prison;
```
ultimaweapon, line 17: Better to make this private and re-export instead or change the other sub-modules to public and remove re-exports.

src/kernel/src/ucred/mod.rs
```diff
@@ -1,25 +1,40 @@
...
 /// An implementation of `ucred` structure.
 #[derive(Debug, Clone)]
 pub struct Ucred {
-    effective_uid: Uid, // cr_uid
-    real_uid: Uid,      // cr_ruid
-    groups: Vec<Gid>,   // cr_groups + cr_ngroups
+    effective_uid: Uid,      // cr_uid
+    real_uid: Uid,           // cr_ruid
+    groups: Vec<Gid>,        // cr_groups + cr_ngroups
+    prison: &'static Prison, // cr_prison
```
ultimaweapon, line 26: How the prison is specified if the process is jailed? If it is allocated a `Prison` then using a static reference here is not correct.

ultimaweapon (changes requested):

SuchAFuriousDeath:
I used lazy_static to fix the prison, if that's fine

src/kernel/src/ucred/mod.rs
```diff
@@ -1,25 +1,41 @@
...
 /// An implementation of `ucred` structure.
 #[derive(Debug, Clone)]
 pub struct Ucred {
-    effective_uid: Uid, // cr_uid
-    real_uid: Uid,      // cr_ruid
-    groups: Vec<Gid>,   // cr_groups + cr_ngroups
+    effective_uid: Uid,  // cr_uid
+    real_uid: Uid,       // cr_ruid
+    groups: Vec<Gid>,    // cr_groups + cr_ngroups
+    prison: Arc<Prison>, // cr_prison
```
ultimaweapon, line 27: Can we try something like `Cow<'static, Prison>` then implement `ToOwned` on `Prison`? If it works we can get rid of `lazy_static`.

ultimaweapon (changes requested):

src/kernel/src/ucred/mod.rs
SuchAFuriousDeath: I didn't think of that at all.

SuchAFuriousDeath: I tried something. Don't know if it's correct

src/kernel/src/ucred/prison.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        false
+    }
+}
+
+impl PartialEq for Prison {
+    fn eq(&self, other: &Self) -> bool {
+        std::ptr::eq(self, other)
+    }
+}
```
ultimaweapon, line 39: Compare a pointer is not really suitable with Rust. We should have some mechanism to generate a unique ID for each `Prison` and use that to identify instead.

src/kernel/src/ucred/mod.rs
```diff
@@ -83,18 +100,31 @@ impl Ucred {
         self.auth.caps.is_unk1() && self.auth.attrs.is_unk2()
     }
 
+    /// See `prison_check` on the PS4 for a reference.
+    pub fn prison_check(&self, other: &Self) -> Result<(), PrisonCheckError> {
```
ultimaweapon, line 104: Can we use the other name that more descriptive? `prison_check` is unclear what checks it does.

src/kernel/src/llvm/mod.rs
```diff
@@ -50,9 +39,10 @@ unsafe impl Sync for Llvm {}
 /// A wrapper on LLVM error.
 #[derive(Debug)]
 pub struct Error {
-    message: String,
+    message: Box<str>,
 }
 
+#[allow(dead_code)]
```
ultimaweapon, line 45: This one should not apply on the implementation.

src/kernel/src/fs/mod.rs
```diff
@@ -33,6 +35,7 @@ mod ioctl;
 mod mount;
 mod path;
 mod perm;
+pub mod socket;
```
ultimaweapon, line 38: This one should not public otherwise we need to make the other modules here public for consistency.

ultimaweapon (changes requested):

src/kernel/src/ucred/mod.rs
```diff
@@ -105,13 +137,49 @@ impl Ucred {
...
+                    todo!()
+                }
+            }
+            _ => {
+                if !pr.allow().intersects(PrisonAllow::ALLOW_SOCKET_AF) {
+                    return Err(PrisonCheckAfError::SocketAddressFamilyNotAllowed(family));
+                }
+            }
+        }
+
```
ultimaweapon, line 177: This one should not be here. It is specific to socket so better to move it there.

src/kernel/src/ucred/mod.rs
```diff
@@ -125,3 +193,31 @@ impl Errno for PrivilegeError {
         }
     }
 }
+
+#[derive(Debug, Error)]
+pub enum PrisonCheckError {
+    #[error("Prison check failed")]
+    CheckFailed,
```
ultimaweapon, line 200: The error is unclear what going on. The message also does not align with Rust guidance.

src/kernel/src/ucred/mod.rs
```diff
@@ -125,3 +193,31 @@ impl Errno for PrivilegeError {
...
+    SocketAddressFamilyNotAllowed(AddressFamily),
+}
+
+impl Errno for PrisonCheckAfError {
+    fn errno(&self) -> NonZeroI32 {
+        match self {
+            Self::SocketAddressFamilyNotAllowed(_) => EAFNOSUPPORT,
+        }
+    }
+}
```
ultimaweapon, line 223: Same as the previous comment.

ultimaweapon (changes requested):

src/kernel/src/ucred/mod.rs
```diff
@@ -83,18 +100,33 @@ impl Ucred {
...
+    pub fn can_see(&self, other: &Self) -> Result<(), PrisonCheckError> {
+        if &self.prison == &other.prison || self.prison.is_child(&other.prison) {
+            return Ok(());
+        }
+
+        Err(PrisonCheckError::InsuffiecientCredentials)
+    }
+
+    pub fn is_jailed(&self) -> bool {
+        self.prison == Cow::Borrowed(&PRISON0)
```
ultimaweapon, line 115: Seems like we use wrong operator here. I think it should be `!=`.

src/kernel/src/ucred/prison.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    pub fn allow(&self) -> PrisonAllow {
+        self.allow
+    }
+
+    pub fn is_child(&self, other: &Self) -> bool {
+        let mut p = other.parent.as_ref();
+
+        while let Some(pr) = p {
+            if self == other {
```
ultimaweapon, line 25: I think this one is not correct.

src/kernel/Cargo.toml
```diff
@@ -9,13 +9,14 @@ path = "src/main.rs"
...
 bitflags = "2.1"
-bytemuck = "1.14.0"
+bytemuck = { version = "1.14.0" , features = ["derive"]}
 byteorder = "1.4"
 clap = { version = "4.1", features = ["derive"] }
 discord-rich-presence = "0.2"
 elf = { path = "../elf" }
 gmtx = { path = "../gmtx" }
 iced-x86 = { version = "1.18", features = ["code_asm"] }
+lazy_static = "1.4.0"
```
ultimaweapon, line 19: I think we don't need this anymore.

ultimaweapon (changes requested):

src/kernel/src/ucred/prison.rs
SuchAFuriousDeath: image

ultimaweapon: What you did here is always checking if `self` is equal to `other` every loop.

SuchAFuriousDeath: OH :skull:

src/kernel/src/fs/mod.rs
```diff
@@ -405,6 +410,91 @@ impl Fs {
...
+
+                Ok(ty)
+            },
+            VFileFlags::FWRITE | VFileFlags::FREAD,
+            &SOCKET_FILEOPS,
+            budget,
+        )?;
+
+        Ok(fd.into())
+    }
```
ultimaweapon, line 453: Can we move this to `net` instead?

src/kernel/src/fs/mod.rs
```diff
@@ -405,6 +410,91 @@ impl Fs {
...
```
ultimaweapon, line 496: Same here.

src/kernel/src/budget/mod.rs
```diff
@@ -57,6 +59,21 @@ impl BudgetManager {
...
+    CpuSet = 4,
+    FdFile = 5,
+    FdSocket = 6,
+    FdEqueue = 7,
+    FdPipe = 8,
+    FdDevice = 9,
+    Threads = 10,
+    FdIpcSocket = 11,
+}
+
```
ultimaweapon, line 76: Where did you get this from?

ultimaweapon (changes requested):

src/kernel/src/budget/mod.rs
SuchAFuriousDeath: image
image
It's in budget_resource_use

SuchAFuriousDeath:
@ultimaweapon I did a big refactor of this and removed a lot of things to make this easier to merge

src/kernel/src/budget/mod.rs
ultimaweapon: Thanks!

ultimaweapon:
What is the state of this?

SuchAFuriousDeath:
I'm not sure, the sockets are merged, because I put them in another PR. I guess I'll close this and put the jail into a separate PR, there's too many merge conflicts. But before that I have to know, is ArcOrStatic a good idea?

ultimaweapon:
> ArcOrStatic

What is that?

SuchAFuriousDeath:
image

SuchAFuriousDeath:
This is how I solved the problem with prisons. I think Cow<'static, Prison> is ... weird. Because we don't want to force ToOwned, because we don't want to clone prisons and then have to compare them with ids... I think.

ultimaweapon:
I agree how the prison is currently working is not good. But I also don't think the `StaticOrArc` is a good idea. Maybe we should create an issue for this so we can find a better solution.

SuchAFuriousDeath:
I mean... will we even have a non-static prison? Aren't we thinking ahead too much?

ultimaweapon:
IIRC all non-prison0 is created dynamically at runtime.

SuchAFuriousDeath:
But we're not emulating the fullkernel just a part of the lifetime, so for us, the the prison the game process is jailed in might be static anyway, no?

ultimaweapon:
> But we're not emulating the full kernel

It is likely we are going to be a real kernel in near future. Actually what I was planned for Obliteration on a far future is it will be an a full fledged OS like FreeBSD or Linux but focus on gaming that people can install on any hardware.

SuchAFuriousDeath:
:flushed:

SuchAFuriousDeath:
Interesting idea. But it better play some games first :D

ultimaweapon:
That is our first goal. But we need keep this idea in mind so we can implement the code that will be benefit this in the future.

SuchAFuriousDeath:
I'm seeing /mnt/sandbox/pfsmnt/{}-app0-patch0-union in the mount function and I'm wondering... will we need pfs too before we can read anything a file :sleepy:?

SuchAFuriousDeath:
If we ever get to an OS level, do you think it would be possible to run Windows executables? That would be pretty insane.

ultimaweapon:
> I'm seeing '/mnt/sandbox/pfsmnt/{}-app0-patch0-union' in the mount function and I'm wondering... will we need pfs too before we can read a file 😪?

We don't need it, for now. Just use HostFS to mock it the same as exfatfs.

> If we ever get to an OS level, do you think it would be possible to run Windows executables? That would be pretty insane.

We can collaborate with Wine so we can provide the API they need from the kernel level.

SuchAFuriousDeath:
> We can collaborate with Wine so we can provide the API they need from the kernel level.
That would be really cool. I recently read an article that they want to implement something as a device in /dev, because emulating some specific functionality is very expensive in userspace, so they want to do it inside the Linux kernel itself.

PR #563, opened 2024-01-20
Reworks limits

SuchAFuriousDeath:


SuchAFuriousDeath:
@ultimaweapon After 500 attempts at fixing the MacOS build, it's finally fixed.

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,75 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum ResourceType {
+    Cpu = 0,
+    Fsize = 1,
+    Data = 2,
+}
+
+impl ResourceType {
+    #[cfg(target_os = "linux")]
+    pub fn into_unix(self) -> libc::__rlimit_resource_t {
```
ultimaweapon, line 14: `into_unix` is quite confused because PS4 is also *nix.

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,75 @@
...
+        match self {
+            Self::Cpu => libc::RLIMIT_CPU,
+            Self::Fsize => libc::RLIMIT_FSIZE,
+            Self::Data => libc::RLIMIT_DATA,
+        }
+    }
+
+    #[cfg(target_os = "macos")]
+    pub fn into_unix(self) -> libc::c_int {
```
ultimaweapon, line 23: Same here.

ultimaweapon (changes requested):

src/kernel/src/process/rlimit.rs
SuchAFuriousDeath: So... 'into_host'?

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,72 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub(super) struct Limits {
+    inner: [ResourceLimit; Self::NLIMITS],
+}
+
+impl Limits {
+    pub const NLIMITS: usize = 3;
```
ultimaweapon, line 35: This one is quite dangerous because if it is out of sync with `ResourceType` we will hit undefined behavior in the `Index<ResourceType>` implementation.

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,72 @@
...
+//TODO: add remaining limits
+#[derive(Debug, Clone, Copy)]
+pub enum ResourceType {
+    Cpu = 0,
+    Fsize = 1,
+    Data = 2,
+}
+
+#[cfg(target_os = "linux")]
+type RawResourceType = libc::__rlimit_resource_t;
```
ultimaweapon, line 13: I think something like `HostResourceType` is better.

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,72 @@
...
+
+#[cfg(target_os = "macos")]
+type RawResourceType = libc::c_int;
```
ultimaweapon, line 16: Same here.

ultimaweapon (changes requested):

src/kernel/src/process/rlimit.rs
ultimaweapon: @SuchAFuriousDeath seems like you are missed this one.

SuchAFuriousDeath: I changed it to get(t as usize).unwrap, so no UB anymore

ultimaweapon: Do you know is it possible to get the maximum value of the enum type? If that possible it will be much more elegance than this one.

SuchAFuriousDeath: I looked, but it's nightly-only :(
https://doc.rust-lang.org/std/mem/fn.variant_count.html

ultimaweapon: I got an idea but you don't need to implement it if you are lazy. The idea is create a derive macro to find out the maximum value of the variants then generate an implementation something like:

```rust
impl ResourceType {
    pub const MAX: usize = 2;
}
```

SuchAFuriousDeath: Unfortunately, I am lazy :( I am playing around with slint though.

ultimaweapon: Okay no problem. We will go with the current solution for now.

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,72 @@
...
+
+        Ok(Self { inner })
+    }
+}
+
+impl Index<ResourceType> for Limits {
+    type Output = ResourceLimit;
+
+    fn index(&self, ty: ResourceType) -> &Self::Output {
+        unsafe { self.inner.get(ty as usize).unwrap() }
```
ultimaweapon, line 55: Now `unsafe` is not needed here.

src/kernel/src/process/rlimit.rs
```diff
@@ -1,4 +1,72 @@
...
+            Self::Cpu => libc::RLIMIT_CPU,
+            Self::Fsize => libc::RLIMIT_FSIZE,
+            Self::Data => libc::RLIMIT_DATA,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(super) struct Limits {
+    inner: [ResourceLimit; Self::NLIMITS],
```
ultimaweapon, line 31: Better to make this struct as a tuple struct instead.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #564, opened 2024-01-20
Revises FS mount

ultimaweapon:
So it will be more idiomatic Rust than C.

### DIFF
--- a/src/kernel/src/fs/dev/mod.rs
+++ b/src/kernel/src/fs/dev/mod.rs
@@ -1,7 +1,9 @@
 pub use self::cdev::*;
 use self::dirent::Dirent;
 use self::vnode::{CHARACTER_OPS, VNODE_OPS};
-use super::{path_contains, DirentType, FsOps, Mode, Mount, MountFlags, Vnode, VnodeType};
+use super::{
+    path_contains, DirentType, FsConfig, FsOps, Mode, Mount, MountFlags, VPathBuf, Vnode, VnodeType,
+};
 use crate::errno::{Errno, EEXIST, ENOENT, EOPNOTSUPP};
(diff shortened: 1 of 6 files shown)


SuchAFuriousDeath:
I actually think the MountOps should be an enum like
`rust
enum MountOpts {
HostFs(HostFsMountOpts),
Devfs(DevfsMountOpts)
...
}
` 
Having it this way would make it impossible to forget an option or provide one that isn't accepted by the FS. Also, we could get rid of all the unwraps, readability +1000. What do you think?

ultimaweapon:
> I actually think the MountOps should be an enum like `enum MountOpts { HostFs(HostFsMountOpts), Devfs(DevfsMountOpts) ... }` Having it this way would make it impossible to forget an option or provide one that isn't accepted by the FS. Also, we could get rid of all the unwraps, readability +1000. What do you think? I can refactor it in my PR.

The problem is mount options usually passed from the user as a string. That mean we need some mechanism to parse it from the string. I think your current solution is nullfs is good enough unless we find a better than that.

SuchAFuriousDeath:
Oh, ok. The mount data could easily be an enum though, right?

SuchAFuriousDeath (approved):

ultimaweapon:
> Oh, ok. The mount data could easily be an enum though, right?

It could but I don't think it is better than `Any` in this case. Trait object should be the right solution here.

ultimaweapon:
Something like this:

```rust
struct Mount {
    data: Arch<dyn Filesystem>,
}

trait Filesystem {}
```

ultimaweapon:
Actually I already tried something like that with this PR but found out it is a massive changes so I decided to revert it otherwise you are going to have a big problem with your nullfs PR.

PR #566, opened 2024-01-21
Replaces VopVector with VnodeBackend

ultimaweapon:
So it is much more idiomatic to Rust. The benefits we got from this is we no longer need to cast `v_data` to the FS specific data.

### DIFF
--- a/src/kernel/src/fs/dev/dirent.rs
+++ b/src/kernel/src/fs/dev/dirent.rs
@@ -7,6 +7,7 @@ use std::sync::{Arc, Weak};
 use std::time::SystemTime;
 
 /// An implementation of `devfs_dirent` structure.
+#[derive(Debug)]
 pub struct Dirent {
     inode: i32,                        // de_inode
     uid: Gutex<Uid>,                   // de_uid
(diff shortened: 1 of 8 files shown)


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Hmmm. won't we need to call the default methods separately from other parts of the code? They should probably be standalone methods and the default implementation in the trait should call those methods.

SuchAFuriousDeath:
If you merge main into this, I'll merge it and deal with conflicts in my branches.

ultimaweapon:
> Hmmm. won't we need to call the default methods separately from other parts of the code? They should probably be standalone methods and the default implementation in the trait should call those methods.

What are the benefits of that?

ultimaweapon:
Sorry I missed the first part. It is unlikely that we need to call the default implementation outside. If that happens we can extract it later. For now it make the code hard to follow without any benefits.

SuchAFuriousDeath:
Yeah I think you're right. I confused the default ops with the handlers in vnodeop_desc

PR #573, opened 2024-01-26
Implement cn_nameiop equivalent

SuchAFuriousDeath:


SuchAFuriousDeath:
@ultimaweapon This is roughly what I had in mind.

src/kernel/src/fs/path.rs
```diff
@@ -286,11 +287,61 @@ impl From<VPathBuf> for String {
...
+    Lookup,
+    Create,
+    Delete,
+    Rename,
+}
+
+/// Inspired by [`std::path::Component`].
+/// Is not an equivalent of the FreeBSD type `componentname`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ComponentName<'a> {
```
ultimaweapon, line 301: `VPathComponent` should be a better name.

src/kernel/src/fs/mod.rs
```diff
@@ -152,14 +152,25 @@ impl Fs {
...
+        &self,
+        path: impl AsRef<VPath>,
+        td: Option<&VThread>,
+    ) -> Result<Arc<Vnode>, LookupError> {
+        self.namei(path, NameiOp::Lookup, td)
+    }
+
     /// This method will **not** follow the last component if it is a mount point or a link.
-    pub fn lookup<P: AsRef<VPath>>(
+    fn namei(
```
ultimaweapon, line 170: I'm not really a fan of `namei` because it make no sense.

src/kernel/src/fs/mod.rs
```diff
@@ -152,14 +152,25 @@ impl Fs {
...
     }
 
+    pub fn lookup(
+        &self,
+        path: impl AsRef<VPath>,
+        td: Option<&VThread>,
+    ) -> Result<Arc<Vnode>, LookupError> {
+        self.namei(path, NameiOp::Lookup, td)
+    }
+
```
ultimaweapon, line 168: We probably don't need this due to my previous comment.

ultimaweapon (changes requested):

src/kernel/src/fs/path.rs
```diff
@@ -286,11 +287,60 @@ impl From<VPathBuf> for String {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LookupOp {
```
ultimaweapon, line 291: This one should not be here.

src/kernel/src/fs/path.rs
```diff
@@ -286,11 +287,60 @@ impl From<VPathBuf> for String {
...
+    Rename,
+}
+
+/// Inspired by [`std::path::Component`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VPathComponent<'a> {
+    Root,
+    Dot,
+    DotDot,
+    Normal { name: &'a str },
```
ultimaweapon, line 304: I think `Name(&'a str)` looks better. It is also much easier to work with unnamed field than named field.

ultimaweapon (changes requested):

src/kernel/src/fs/path.rs
ultimaweapon: Seems like to missed this one. It should be in the root of `fs` instead of `path`.

PR #578, opened 2024-01-27
Moves deci device from TtyManager to DebugManager

ultimaweapon:


src/kernel/src/debug/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        _: i32,
+        _: Option<&VThread>,
+    ) -> Result<(), Box<dyn Errno>> {
+        todo!()
+    }
+}
+
+/// Represents an error when [`DebugManager`] is failed to initialize.
+#[derive(Debug, Error)]
+pub enum DebugManagerError {
```
SuchAFuriousDeath, line 60: I would rename this to something like DebugManagerInitError if this will only be used in the 'new' function

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath (approved):

SuchAFuriousDeath (approved):

PR #581, opened 2024-01-28
implement syscall 121 sys_writev

mikusp:


src/kernel/src/fs/mod.rs
```diff
@@ -272,6 +273,27 @@ impl Fs {
...
+        let file = td.proc().files().get(fd).ok_or(SysErr::Raw(EBADF))?;
+        let mut written = 0;
+        for i in 0..len - 1 {
+            let buf: &[u8] = unsafe {
+                std::slice::from_raw_parts(
+                    ptr.offset(i.try_into().unwrap()).read().iov_base as *const u8,
+                    ptr.offset(i.try_into().unwrap()).read().iov_len,
+                )
+            };
+            written = written + file.write(buf as _, Some(&td))?;
```
SuchAFuriousDeath, line 291: I just looked at this today and I don't think this is correct. Currently, our read and write-related functions don't align with FreeBSD, i.e. file.write
doesn't take an iovec, but just a regular slice, it will have to be refactored.

src/kernel/src/fs/mod.rs
```diff
@@ -272,6 +273,27 @@ impl Fs {
         Ok(written.into())
     }
 
+    fn sys_writev(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
+        let fd: i32 = i.args[0].try_into().unwrap();
+        let ptr: *const libc::iovec = i.args[1].into();
+        let len: usize = i.args[2].try_into().unwrap();
+
+        let td = VThread::current().unwrap();
+        let file = td.proc().files().get(fd).ok_or(SysErr::Raw(EBADF))?;
```
SuchAFuriousDeath, line 282: This also isn't that simple. I assume you got this from sys_write, but it's incomplete there. It might be ok for now though, we'll have to wait for ultima.

SuchAFuriousDeath (changes requested):

src/kernel/src/fs/mod.rs
```diff
@@ -272,6 +273,27 @@ impl Fs {
         Ok(written.into())
     }
 
+    fn sys_writev(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
+        let fd: i32 = i.args[0].try_into().unwrap();
+        let ptr: *const libc::iovec = i.args[1].into();
```
SuchAFuriousDeath, line 278: Yeah and furthermore, there is no libc on Windows, this should be our own type. And in general, I think we should use our own types in the kernel and only use things like libc when communicating directly with the host system

SuchAFuriousDeath (changes requested):

mikusp:
Thanks for the review, I'll take a closer look at what ps4 does

SuchAFuriousDeath:
In this case I think it does exactly what FreeBSD does. You can look at it here, for example:
http://fxr.watson.org/fxr/source/kern/sys_generic.c?v=FREEBSD-9-STABLE#L455

src/kernel/src/fs/mod.rs
SuchAFuriousDeath: @ultimaweapon What do you think about this?

ultimaweapon: Do you mean this line? If yes I don't have any problem with it. But I did not take a look on the PS4 so I can't tell if it is correct.

SuchAFuriousDeath: It should be fget_write, which does some additional checks and sets some flags. This one line is a 'weak' version of fget_write

ultimaweapon: @SuchAFuriousDeath can you create an issue for this this so we don't forget?

SuchAFuriousDeath:
@mikusp I created an issue where I described more closely what has to be done: #587.
If you want to, you can do all the write syscalls at the same time because they are basically all the same.

src/kernel/src/fs/mod.rs
SuchAFuriousDeath: Actually, looking at this now, it's less incorrect that I thought, you call file.write in a loop. But it should still be redone using vectored writes.

src/kernel/src/fs/mod.rs
```diff
@@ -272,6 +273,27 @@ impl Fs {
...
+        let len: usize = i.args[2].try_into().unwrap();
+
+        let td = VThread::current().unwrap();
+        let file = td.proc().files().get(fd).ok_or(SysErr::Raw(EBADF))?;
+        let mut written = 0;
+        for i in 0..len - 1 {
+            let buf: &[u8] = unsafe {
+                std::slice::from_raw_parts(
```
SuchAFuriousDeath, line 286: Also, it doesn't really make sense to create an iovec(which is basically a slice) and turn it back into a slice again. 
But this will have to be changed during the refactor.

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath:
@ultimaweapon By the way, I think VFile can't implement Seek, Write and Read, they all require mutable references :no_mouth:

ultimaweapon:
> @ultimaweapon By the way, I don't think VFile can implement Seek, Write and Read, they all require mutable references 😶

We can put it behind a mutex.

PR #582, opened 2024-01-28
sys_sigaction with oact != NULL

mikusp:


src/kernel/src/process/signal.rs
```diff
@@ -73,4 +77,26 @@ impl SignalActs {
     pub fn remove_catch(&mut self, sig: NonZeroI32) {
         self.catch.remove(sig);
     }
+
+    pub fn flags(&self, sig: NonZeroI32) -> SignalFlags {
+        let mut flags: SignalFlags = SignalFlags::ZERO;
```
SuchAFuriousDeath, line 82: This should be SignalFlags::empty().

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath:
image
What about this bit? When I mentioned flags, I meant the rest of the branch too :D It seems like it's the same with FreeBSD:
http://fxr.watson.org/fxr/source/kern/kern_sig.c?v=FREEBSD-9-STABLE#L665

SuchAFuriousDeath (approved):

PR #583, opened 2024-01-28
mmap with MAP_FIXED

mikusp:


src/kernel/src/memory/mod.rs
```diff
@@ -147,7 +151,18 @@ impl MemoryManager {
...
+            let pageoff = offset & 0x3fff;
+            let adjusted_addr = addr - pageoff;
+            let adjusted_len = len + pageoff;
+            let rounded_len = Self::round_page(adjusted_len);
+
+            if (adjusted_addr & 0x3fff) != 0 {
+                return Err(MmapError::InvalidOffset);
+            }
+            // TODO: check if addr is within min and max mappings for this proc
+            if adjusted_addr + rounded_len < adjusted_addr {
```
ultimaweapon, line 163: Where is `adjusted_addr + rounded_len` come from? What I saw on the PS4 is it negated the `adjusted_addr` instead.

ultimaweapon (changes requested):

Issue #592, opened 2024-01-31
Make the Errno trait derivable


SuchAFuriousDeath:

SuchAFuriousDeath:
Technically, it's not fully derivable, that's why I didn't close the issue. Whatever :D

PR #593, opened 2024-02-01
Makes errno derivable

SuchAFuriousDeath:
Progresses #592.

### DIFF
--- a/src/kernel/src/dmem/blockpool.rs
+++ b/src/kernel/src/dmem/blockpool.rs
@@ -1,7 +1,7 @@
 use crate::errno::{Errno, ENOTTY, ENXIO};
 use crate::fs::{IoCmd, VFile, VFileOps};
 use crate::process::VThread;
-use std::num::NonZeroI32;
+use macros::Errno;
 use thiserror::Error;
 
 const BLOCKPOOL_FILEOPS: VFileOps = VFileOps {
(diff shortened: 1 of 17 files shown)


SuchAFuriousDeath:
The question is, how do we deal with variants that have a nested error? Do we put #[errno(auto)], or something like that? I could take the #[from] and #[source] attribute from the Error proc macro, but I'm not sure if we want to do it that way, since not every nested error variant necessarily has to have its error be e.errno(), it could be a static one as well and you could forget.

ultimaweapon:
> The question is, how do we deal with variants that have a nested error? Do we put #[errno(auto)], or something like that? I could take the #[from] and #[source] attribute from the Error proc macro, but I'm not sure if we want to do it that way, since not every nested error variant necessarily has to have its error be e.errno(), it could be a static one as well and you could forget.

If we force the error type to be an enum it is impossible to forget otherwise it will be a compilation error.

SuchAFuriousDeath:
I don't really see how that's true. Look at this for example:
image
The inner error of the GetVnodeFailed variant is ignored and instead it's mapped to EIO. You could forget that and it wouldn't be a compiler error.

VocalFan (approved):

VocalFan (approved):

PR #594, opened 2024-02-02
Makes nullfs usable

SuchAFuriousDeath:


ultimaweapon:
> I'm still not quite sure how to mount nullfs properly, since it has to be mounted on a vnode, but we don't currently have the games directory mounted. I guess we'll have to do another host mount?

My idea is to implement a fake PFS and use it to mount a game directory as a fake blue ray disc.

SuchAFuriousDeath:
So... I should work on pfs next?

ultimaweapon:
I think just make HostFS configurable so it can pretend to be any FS is better.

BTW I think we need more people so your PRs can be reviewed faster because I can take a look only on weekends. Some weekends I can't even do it because I have something to do. The problem is how to find more people?

SuchAFuriousDeath:
I'm not sure. This one is pretty safe to merge I think.

SuchAFuriousDeath:
By the way, the new guy making the PRs, mikusp, told me in Discord that his PRs aren't random, but that they implement things he himself encountered (He managed to somehow bypass what we're currently stuck on)

SuchAFuriousDeath:
@ultimaweapon Here is his hacky WIP branch: https://github.com/mikusp/obliteration/commit/e5e4df13b839120c47933e761429961341cb7026. It's actually pretty insane how far he got. I suggest we let him into the org.

SuchAFuriousDeath:
@ultimaweapon The ?Sized bound seems really weird here btw.
image
Either it should be on the trait itself(which applies it everywhere), or nowhere

VocalFan (approved):

ultimaweapon:
@mikusp wanna join the @obhq organization? We need more developers so the PR can be reviewed properly.

mikusp:
Yes of course!

ultimaweapon:
Invitation has been sent.

PR #604, opened 2024-02-03
#603 Implement sys_nanosleep

mikusp:
I've included the Windows implementation, but I have no idea what am I doing there so there might be bugs.

This is not 100% correct with PS4, as it will always sleep the requested amount on both Windows and Linux - that should not be a problem though as libkernel functions would have restarted on EINTR anyway

### DIFF
--- a/src/kernel/src/time/mod.rs
+++ b/src/kernel/src/time/mod.rs
@@ -1,9 +1,11 @@
-use crate::errno::Errno;
+use crate::errno::{Errno, EINVAL};
+use crate::info;
 use crate::syscalls::{SysErr, SysIn, SysOut, Syscalls};
+use std::io::Error;
+use std::mem::zeroed;
 use std::num::NonZeroI32;
 use std::sync::Arc;
 use thiserror::Error;
-
 pub struct TimeManager {}
 
 impl TimeManager {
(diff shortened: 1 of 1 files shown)


SuchAFuriousDeath:
@mikusp Doesn't this sleep the thread running the kernel? We want to sleep the thread that's requesting that's making the syscall, don't we?

mikusp:
I believe nanosleep would sleep the calling thread, so should be correct

mikusp:
I've rebased the branch and fixed the windows implementation so it compiles at least

src/kernel/src/process/sleep.rs
```diff
@@ -0,0 +1,95 @@
+use crate::syscalls::{SysErr, SysOut};
+use std::{io::Error, num::NonZeroI32};
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct Timespec {
+    pub seconds: i64,     // tv_sec
+    pub nanoseconds: i64, // tv_nsec
+}
```
ultimaweapon, line 9: I think this one should not be here. My idea is create a new module `time` on the top-level and move it to that module.

ultimaweapon (dismissed):

mikusp:
Updated now with the nanosleep function moved to the time module

PR #605, opened 2024-02-03
Implement /dev/sbl_srv file

mikusp:
If there is a way to avoid making `dev` mod pub, let me know in suggestions.

### DIFF
--- a/src/kernel/src/fs/mod.rs
+++ b/src/kernel/src/fs/mod.rs
@@ -25,7 +25,7 @@ pub use self::perm::*;
 pub use self::stat::*;
 pub use self::vnode::*;
 
-mod dev;
+pub mod dev;
 mod dirent;
 mod file;
 mod host;
--- a/src/kernel/src/sbl/mod.rs
+++ b/src/kernel/src/sbl/mod.rs
@@ -1,0 +1,45 @@
+use std::sync::Arc;
+
+use crate::{
+    errno::Errno,
+    fs::{make_dev, Cdev, CdevSw, DriverFlags, MakeDev, Mode, OpenFlags},
+    process::VThread,
+    ucred::{Gid, Uid},
+};
+
+#[derive(Debug)]
+pub struct SblManager {}
+
+impl SblManager {
+    pub fn new() -> Arc<Self> {
+        let sbl = Arc::new(Self {});
+
+        let sbl_srv_cdevsw = Arc::new(CdevSw::new(
+            DriverFlags::D_INIT,
+            Some(Self::sbl_srv_open),
+            None,
+        ));
+
+        let _ = make_dev(
+            &sbl_srv_cdevsw,
+            0,
+            "sbl_srv",
+            Uid::ROOT,
+            Gid::ROOT,
+            Mode::new(0o600).unwrap(),
+            None,
+            MakeDev::MAKEDEV_ETERNAL,
+        );
+
+        sbl
+    }
+
+    fn sbl_srv_open(
+        _: &Arc<Cdev>,
+        _: OpenFlags,
+        _: i32,
+        _: Option<&VThread>,
+    ) -> Result<(), Box<dyn Errno>> {
+        todo!()
+    }
+}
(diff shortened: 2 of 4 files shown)


SuchAFuriousDeath:
We should probably initialize all the /dev devices somewhere inside the module.

src/kernel/src/fs/dev/sblsrv.rs
```diff
@@ -0,0 +1,41 @@
...
+/// An implementation of `/dev/sbl_srv`.
+#[derive(Debug)]
+pub struct SblService {}
+
+impl SblService {
+    pub const PATH: &'static VPath = vpath!("/dev/sbl_srv");
+
+    pub fn new() -> Arc<Self> {
+        let sbl_srv_cdevsw = Arc::new(CdevSw::new(
+            DriverFlags::from_bits_retain(0x80000000),
```
SuchAFuriousDeath, line 18: This is DriverFlags::NEEDS_MINOR

SuchAFuriousDeath (changes requested):

src/kernel/src/fs/dev/sblsrv.rs
```diff
@@ -0,0 +1,41 @@
...
+            Some(|_, _, _, _| Ok(())),
```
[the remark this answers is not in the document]
SuchAFuriousDeath: By the way, we are pretty much categorically against stubbing like this. We always try to implement things as much as we can.

ultimaweapon:
> We should probably initialize all the /dev devices somewhere inside the module.

Better to split each device into a separated module like `dev/dmem0` will be a responsible of `DmemManager`.

ultimaweapon:
All of the devices in the `src/kernel/src/fs/dev` are legacy ones. It is a remnant of the old FS system. That means we are going to move all of them to the outside like `TtyManager`.

SuchAFuriousDeath:
@ultimaweapon I think I already asked this somewhere, but you may have overlooked it. Aren't all the Cdevsws static in FreeBSD?

ultimaweapon:
> @ultimaweapon I think I already asked this somewhere, but you may have overlooked it. Aren't all the Cdevsws static in FreeBSD? I don't really see a reason for Arc here

I can't remember why I was chosen for the current solution. But I already thought about static but decided to use Arc instead. Maybe we can try converted to a static version and see if we hit any problems.

src/kernel/src/fs/dev/sblsrv.rs
mikusp: NEEDS_MINOR is 0x00800000, according to FreeBSD that is D_INIT

ultimaweapon:
@SuchAFuriousDeath can you check if your issues has been fixed?

SuchAFuriousDeath:
Depends on whether you're fine with merging this
image

ultimaweapon:
I think @mikusp already fixed that.

SuchAFuriousDeath:
Oh :eyes:

src/kernel/src/fs/mod.rs
```diff
@@ -25,7 +25,7 @@ pub use self::perm::*;
 pub use self::stat::*;
 pub use self::vnode::*;
 
-mod dev;
+pub mod dev;
```
ultimaweapon, line 28: This should not need to be public anymore.

src/kernel/src/sbl/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        let _ = make_dev(
+            &sbl_srv_cdevsw,
+            0,
+            "sbl_srv",
+            Uid::ROOT,
+            Gid::ROOT,
+            Mode::new(0o600).unwrap(),
+            None,
+            MakeDev::MAKEDEV_ETERNAL,
```
ultimaweapon, line 31: Is not this one is zero here?

ultimaweapon (changes requested):

SuchAFuriousDeath:
Superseded by #729

PR #606, opened 2024-02-03
Handle missing app version

mikusp:
I've found some homebrew that was missing app_version and crashing on that unwrap()


ultimaweapon (approved):
Thanks!

PR #609, opened 2024-02-04
initialize vnops implementation

SuchAFuriousDeath:


SuchAFuriousDeath:
@ultimaweapon This is unrelated to this PR, but I think we should make IoCmd and enum. Some of the variants could carry a mutable or immutable reference to something. Currently, we pass the data as a &mut [u8] and cast it using bytemuck, but we'll have to do this every time we want to get a reference. Also, the &mut [u8] carries and extra len which is redundant, because we already have it in the command.

SuchAFuriousDeath:
@ultimaweapon Can you make the filesystem lookup the file correctly? I didn't really understand how you planon doing it and it's difficult to work on syscalls we haven't reached yet.

ultimaweapon:
Will do tomorrow after I finish with my employer work.

SuchAFuriousDeath:
By the way, should all device really be in their own *Manager thing? I think we should probably do the same thing as with Filesystem and VnodeBackend.

ultimaweapon:
> By the way, should all device really be in their own *Manager thing? I think we should probably do the same thing as with Filesystem and VnodeBackend. I like how they are in one place in fs/dev

Does that mean any module that also have a device need to be in `fs` like dmem?

src/kernel/src/fs/file.rs
```diff
@@ -37,6 +41,10 @@ impl VFile {
         &mut self.flags
     }
 
+    pub fn vnode(&self) -> &Arc<Vnode> {
+        todo!()
+    }
+
```
ultimaweapon, line 47: IIRC `f_vnode` can be null for a file that is not backed by a vnode.

ultimaweapon (dismissed):

SuchAFuriousDeath:
> > By the way, should all device really be in their own *Manager thing? I think we should probably do the same thing as with Filesystem and VnodeBackend. I like how they are in one place in fs/dev
> 
> Does that mean any module that also have a device need to be in `fs` like dmem?

Potentially. While defined in fs/dmem, the devices could of course depend on other things defined elsewhere. 
Maybe it's not such a good idea after all :D

SuchAFuriousDeath:
By the way, do you think it's necesarry to stick so close to FreeBSD with the devices? I haven't dug in too deep yet, but it feels like we might not need the complexity, but I'm not sure

mikusp (approved):

SuchAFuriousDeath:
By the way, in FreeBSD, it seems that only one device actually uses fdopen in cdevsw :sweat: I really think we should move towards a more Rusty trait based implementation.

PR #633, opened 2024-02-09
initialize sys_osem_create

SuchAFuriousDeath:


SuchAFuriousDeath:
Excuse me?? 
image
xDDD

src/kernel/src/osem/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+        };
+
+        let mut objects = self.proc.objects_mut();
+
+        let (entry, id) = objects.alloc::<_, ()>(|_| Ok(Osem::new(flags))).unwrap();
+
+        entry.set_name(Some(name.to_owned()));
+        entry.set_ty(0x120);
+
+        todo!()
```
mikusp, line 44: Why end with todo!()? Even if the implementation is not complete, returning anything will allow the program to proceed

mikusp (approved):

src/kernel/src/osem/mod.rs
SuchAFuriousDeath: Yeah, but we don't really want to do things this way. There's some things that still have to be done in this syscall and it's probably better to do them when we reach it

SuchAFuriousDeath: Actually, you're correct. We might as well return right away, because we do return the right value, it's analogous to the kqueue PR. I'll change it after work

mikusp (approved):

PR #636, opened 2024-02-11
Mounts /mnt

ultimaweapon:
@SuchAFuriousDeath here is the idea for what you have asked me how to make FS usable. Basically we need to mount FS the same as PS4. When this is done and all of required FS operations has been implemented everything should be working automatically.


ultimaweapon:
BTW today I don't have much time to work on this and I think this will apply for the next week too. So you can take over the tmpfs if you want.

SuchAFuriousDeath (approved):

PR #648, opened 2024-02-15
Partially implements sys_lseek, refactors ioctl commands

SuchAFuriousDeath:


mikusp (approved):
That macro is above my Rust understanding 😂

SuchAFuriousDeath:
It should probably be a proc macro. By the way you can see the result with cargo expand fs::ioctl.

mikusp (approved):

PR #649, opened 2024-02-16
Initializes sockets

SuchAFuriousDeath:


src/kernel/src/net/socket.rs
```diff
@@ -0,0 +1,131 @@
...
+        todo!()
+    }
+}
+
+#[derive(Debug, Error)]
+enum ReceiveError {}
+
+impl Errno for ReceiveError {
+    fn errno(&self) -> NonZeroI32 {
+        todo!()
```
mikusp, line 115: What's the reasoning for using todo!() here and match *self {} in previous empty enum?

mikusp (approved):

src/kernel/src/net/socket.rs
SuchAFuriousDeath: The error has to implement the Errno trait so that it can be into()'d into Box<dyn Errno> in the read function.

PR #650, opened 2024-02-16
regmgr: decode_key with v3 == 1

mikusp:
Computing x uses SBOX1 in reverse, and I see no other way in Rust to index from the end? .reverse() reverses in-place


SuchAFuriousDeath:
You could an extension trait for getting a reversed clone of an array, maybe it could even be const.

mikusp:
Maybe sth like that? https://stackoverflow.com/a/77674505

SuchAFuriousDeath:
Sure, why not

SuchAFuriousDeath (approved):

PR #655, opened 2024-02-16
Handle R_X86_64_DTPOFF64

mikusp:
Created as draft, as I don't know what I'm doing there. DTPOFF64 seems similar to R_X86_64_64, but without including the base address. I'm not sure if that's a correct way or better to reuse Self::get_relocated?


src/kernel/src/rtld/mod.rs
```diff
@@ -792,6 +792,17 @@ impl<E: ExecutionEngine> RuntimeLinker<E> {
...
+                    let md = match resolver.resolve_with_local(md, sym, symflags) {
+                        Some((md, _)) => md,
+                        None => continue,
+                    };
+
+                    let sym = md.symbol(sym).unwrap();
+                    let addr = sym.value().wrapping_add_signed(addend);
+
+                    (Relocated::Data((md, addr)), addr)
+                }
```
ultimaweapon, line 805: What I saw is it also read the value from the target address and add some values to it.

ultimaweapon (changes requested):

src/kernel/src/rtld/mod.rs
red-prig: data:=(data + rela^.r_addend + def^.st_value);

ultimaweapon (approved):

SuchAFuriousDeath:
I merged main into this, so you should be able to merge this now.

PR #656, opened 2024-02-17
vm.budgets.mlock_avail name2oid

mikusp:


SuchAFuriousDeath:
You forgot to draw it into the tree :(

mikusp:
On the second look I'm not sure if I've read the tree correctly 🤔 It looks in ghidra like 314 is a child of 313 but judging from other sysctls that feels wrong 😓

src/kernel/src/sysctl/mod.rs
```diff
@@ -960,6 +973,38 @@ static VM_PS4DEV_TRCMEM_AVAIL: Oid = Oid {
     enabled: true,
 };
 
+static VM_BUDGETS: Oid = Oid {
+    parent: &VM_CHILDREN,
+    link: None, // TODO: Change to a proper value.
+    number: Sysctl::VM_BUDGETS,
+    kind: Sysctl::CTLFLAG_RD | Sysctl::CTLFLAG_WR | Sysctl::CTLTYPE_NODE,
```
SuchAFuriousDeath, line 980: Use Sysctl::CTLFLAG_RW instead of the first two

src/kernel/src/sysctl/mod.rs
```diff
@@ -960,6 +973,38 @@ static VM_PS4DEV_TRCMEM_AVAIL: Oid = Oid {
...
+    arg2: 0,
+    name: "budgets",
+    handler: None,
+    fmt: "N",
+    descr: "VM budgets",
+    enabled: true,
+};
+
+static VM_BUDGETS_CHILDREN: OidList = OidList {
+    first: Some(&VM_BUDGETS_MLOCK_AVAIL), // TODO: use a proper value
```
SuchAFuriousDeath, line 991: Isn't it already the proper value? :thinking:

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath (approved):

PR #657, opened 2024-02-17
Device trait

SuchAFuriousDeath:


src/kernel/src/fs/dev/cdev.rs
```diff
@@ -149,3 +201,40 @@ bitflags! {
 pub type CdevOpen = fn(&Arc<Cdev>, OpenFlags, i32, Option<&VThread>) -> Result<(), Box<dyn Errno>>;
 pub type CdevFd =
     fn(&Arc<Cdev>, OpenFlags, Option<&VThread>, Option<&mut VFile>) -> Result<(), Box<dyn Errno>>;
+
+/// An implementation of the `cdevsw` structure.
+pub(super) trait Device: Debug + Sync + Send + 'static {
```
mikusp, line 206: Is the old one above obsolete now?

mikusp (approved):

src/kernel/src/fs/dev/cdev.rs
SuchAFuriousDeath: it will be eventually, but I don't want to refactor it without ultima, because this will be a pretty big change

PR #661, opened 2024-02-17
refactors idt

SuchAFuriousDeath:


SuchAFuriousDeath:
I think VnodeBackend::open should return a VFileType

src/kernel/src/namedobj/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#[derive(Debug)]
+pub struct NamedObj {
+    name: String,
+    data: usize,
+}
+
+impl NamedObj {
+    pub fn new(name: impl ToString, data: usize) -> Self {
```
ultimaweapon, line 58: Better to use `Into<String>` instead so it does not allocate a new string if `name` is a `String`.

ultimaweapon (changes requested):

mikusp (dismissed):

mikusp (approved):

SuchAFuriousDeath:
> I think VnodeBackend::open should return a VFileType btw

@ultimaweapon Because of this: 
image
Also, I'm dumb and realized just now that f_vnode is set here in kern_open:
image

Issue #48, opened 2022-11-18
The Kernel Crashed.


VocalFan:
When trying to run Amplitude (And yes, I did a fresh install), Kyty would usually give a missing instruction error. However, when running the game, it now gives a "The kernel crashed" error before Obliteration closes entirely. It doesn't even show the Kyty error.

![image](https://user-images.githubusercontent.com/45863583/202610268-409688a5-0f9d-4c48-ba8b-1c1548ce3dd4.png)
![image](https://user-images.githubusercontent.com/45863583/202610374-f3a6e35f-f65c-4727-9a23-7d96e160fd43.png)
![image](https://user-images.githubusercontent.com/45863583/202610292-8a3f158c-b10d-480d-85fa-6b7ec497331d.png)

ultimaweapon:
Strange. Recently changes have nothing to do with the kernel. If Rust was panic it should tell it was panic but nothing was printed.

Mou-Ikkai:
Any possible way I can help further?

ultimaweapon:
Are you able to build on your machine? If yes try using that build to see if this problem still happens.

ultimaweapon:
Try this build https://github.com/ultimaweapon/obliteration/actions/runs/3514045837 to see if it works.

Mou-Ikkai:
Latest builds have fixed the crash.

Mou-Ikkai:
I'm saying before your github action changes. The "Honors ActiveFat" change fixed it.

ultimaweapon:
Hornors ActiveFat does not fix this issue. The reason you see it working was because the cached object files just accidentally worked with the installed toolchain on the runner. That means sometimes it will work and sometimes it will not.

Mou-Ikkai:
I'll investigate this :3

ultimaweapon:
Any updates? If you can't find other reasons for this problem I will reset main branch back to my commit.

Mou-Ikkai:
> Any updates? If you can't find other reasons for this problem I will reset main branch back to my commit.

Continued testing... Every build after that one is flawless... Checking the files, it seems like the kernel file was corrupted :<

ultimaweapon:
You can test by building the problematic commit on your machine without any object files from the previous build (remove `build` and `src/target` directory before building). If it work that mean the problem come from the cached files as I suspected.

Mou-Ikkai:
[...]

PR #663, opened 2024-02-18
Refactors regmgr_call and privileges

SuchAFuriousDeath:


src/kernel/src/regmgr/command.rs
```diff
@@ -0,0 +1,30 @@
+use super::RegError;
+
+#[repr(u32)]
+pub(super) enum RegMgrCommand<'a> {
+    SetInt(&'a SetIntArg) = 0x18,
+    Unk1(&'a Unk1Arg) = 0x19,
+}
+impl RegMgrCommand<'_> {
+    pub fn try_from_raw_parts(cmd: u32, arg: *const u8) -> Result<Self, RegError> {
```
ultimaweapon, line 9: This method should be unsafe and bind a lifetime to `arg` otherwise it is possible for the returned object to outlive the `arg`.

ultimaweapon (changes requested):

src/kernel/src/regmgr/command.rs
SuchAFuriousDeath: Actually, I think it will be better to make it take a SysArg directly

src/kernel/src/fs/ioctl.rs
```diff
@@ -43,15 +45,17 @@ macro_rules! commands {
                 pub const IOC_OUT: u32 = 0x40000000;
                 pub const IOC_IN: u32 = 0x80000000;
 
-                pub fn try_from_raw_parts(cmd: u64, arg: *mut u8) -> Result<Self, SysErr> {
+                /// # Safety
+                /// `arg` has to be a pointer to the correct value
+                pub unsafe fn try_from_raw_parts(cmd: u64, arg: SysArg) -> Result<Self, SysErr> {
```
ultimaweapon, line 50: This is even dangerous than the pointer because it hide the fact that the returned value have a lifetime bind to some pointer.

I guess we should required every PRs that use unsafe to be reviewed by me (or someone else who can catch this kind of bug) so we don't have a dangerous bug get merged. Here is an example of what going to happen if we violate the Rust assumption due to unsafe: https://github.com/obhq/obliteration/issues/48

ultimaweapon (changes requested):

src/kernel/src/fs/ioctl.rs
SuchAFuriousDeath: I tried something, take a look

ultimaweapon: Better to use something like this: https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr

SuchAFuriousDeath: It's that simple? :D

SuchAFuriousDeath: Wait a minute. To be completely honest, sometimes I don't fully understand lifetimes. The CStr struct isn't lifetimed, but our IoCmd and RegMgrCommand are.  Doesn't that make any difference?

ultimaweapon: The point is `try_from_raw_parts` need to have an explicit lifetime and document it how it bind to the pointer.

PR #667, opened 2024-02-18
Do not crash on tid == -1 in sys_thr_set_name

mikusp:
Previous implementation thought of tid == -1 but it crashed on unwrapping the conversion to i32. This is called by system app ShellAppProxy

### DIFF
--- a/src/kernel/src/process/mod.rs
+++ b/src/kernel/src/process/mod.rs
@@ -421,7 +421,7 @@ impl VProc {
     }
 
     fn sys_thr_set_name(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
-        let tid: i32 = i.args[0].try_into().unwrap();
+        let tid: i64 = i.args[0].into();
         let name: Option<&str> = unsafe { i.args[1].to_str(32) }?;
 
         if tid == -1 {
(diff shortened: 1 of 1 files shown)


src/kernel/src/process/mod.rs
```diff
@@ -421,7 +421,7 @@ impl VProc {
     }
 
     fn sys_thr_set_name(self: &Arc<Self>, i: &SysIn) -> Result<SysOut, SysErr> {
-        let tid: i32 = i.args[0].try_into().unwrap();
+        let tid: i64 = i.args[0].try_into().unwrap();
```
SuchAFuriousDeath, line 424: This is an infallible conversion, it should be .into(). You'll have to write a From implementation for this I think

SuchAFuriousDeath (dismissed):

src/kernel/src/process/mod.rs
mikusp: There already is a From instance for that 👍

SuchAFuriousDeath: Oh right, I actually wrote it, but I thought it was in one of the yet unmerged PRs :D

SuchAFuriousDeath (approved):

VocalFan (approved):

SuchAFuriousDeath:
I told VocalFan to merge this, because it's already done this way in sys_cpuset_getaffinity

PR #672, opened 2024-02-18
Implements sys_dynlib_get_info

SuchAFuriousDeath:


SuchAFuriousDeath:
I think this is correct now, but we should probably wait until this syscall is reached until we merge this

SuchAFuriousDeath:
@ultimaweapon By the way, shouldn't relro and relro_size be stored inside a module too?

ultimaweapon:
> @ultimaweapon By the way, shouldn't relro and relro_size be stored inside a module too?

Only if we need to use it somewhere later.

ultimaweapon:
> We do here, that's why I'm asking :D

Then the answer is yes.

SuchAFuriousDeath:
Thanks to prig, we know the data structure more accurately.

src/kernel/src/rtld/module.rs
```diff
@@ -739,14 +758,14 @@ bitflags! {
...
-        const MAIN_PROG = 0x0001;
+        const MAINPROG = 0x0001;
         const TEXT_REL = 0x0002;
         const JMPSLOTS_DONE = 0x0004; // TODO: This seems incorrect.
         const TLS_DONE = 0x0008;
         const INIT_SCANNED = 0x0010;
         const ON_FINI_LIST = 0x0020;
         const DAG_INITED = 0x0040;
-        const UNK1 = 0x0100; // TODO: Rename this.
+        const IS_SYSTEM = 0x0100; // TODO: Rename this.
```
mikusp, line 768: No need for the todo anymore?

mikusp (dismissed):

mikusp (approved):

VocalFan (approved):
Don't kill me for this lmao

PR #674, opened 2024-02-19
Implement sys_thr_get_name

mikusp:


SuchAFuriousDeath:
I'll check this when I have time.
@ultimaweapon This is unrelated to the PR, but what would you thinkabout doing Vthread::current().unwrap() before a syscall is called and making it another argument to all the syscall handlers? I feel like it's kind of redundant to do Vthread::current().unwrap() everywhere when we could do it once. Maybe there is an issue with it that I don't see.

PR #691, opened 2024-02-24
Adds Actions to automatically add/remove S-ready from PRs

ultimaweapon:


VocalFan:
I get the last one being Build-out-of-date, but what is this one?

ultimaweapon:
The previous one has some flaws so we will move to this one instead.

SuchAFuriousDeath:
I've been preliminarily looking at how IPC sockets work, and it looks like it's a looooot of code.

SuchAFuriousDeath:
From the looks of it, sys_connect binds IPC sockets to a filesystem path.

SuchAFuriousDeath:
I think socket protocols and address families could (theoretically) be generalized in a nice way, roughly like this. Hopefully we get to do this one day. 
image

SuchAFuriousDeath:
Actually more like this:
image

ultimaweapon:
It is likely needed to be a trait object so the trait needs to be object safe. BTW better to create a discussion instead of hooking into a random issue/PR so we can easily find it.

SuchAFuriousDeath (approved):

PR #696, opened 2024-02-25
Initializes KVM and WHP

ultimaweapon:


SuchAFuriousDeath:
And also, regarding the previous PR, shouldn't this be a an empty enum?
```
#[repr(C)]
pub struct hv_vm_config_t([u8; 0]);
```
Then, it can never be contructed and can only exist behind a pointer. But I'm not quite sure what your intentions were behind this.

SuchAFuriousDeath:
llvm-sys and libc still use empty enums by the way :D

SuchAFuriousDeath:
Is this left as draft on purpose?

ultimaweapon:
> Is this left as draft on purpose?

Correct.

SuchAFuriousDeath:
By the way, mikusp said on Discord that he reached sys_thr_new

SuchAFuriousDeath (approved):

PR #708, opened 2024-02-27
Refactors VFile::is_seekable to use VFile::vnode() instead

SuchAFuriousDeath:


SuchAFuriousDeath:
@ultimaweapon By the way, I think that the vnode item field should be remove and each corresponding variant should be inside VnodeType.
image
Currently, we do unnecesarry checks and theoretically, we allow nonsense states (Like a device vnode without a device). FreeBSD has to check the vnode type to ensure that v_un is a pointer of the right type.

mikusp (approved):

PR #715, opened 2024-03-02
Initialize send and receive syscalls

SuchAFuriousDeath:


src/kernel/src/fs/uio.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// This is for when the PS4 DOES check the length (such as in recvmsg, recvfrom, sendmsg and sendto)
+    pub unsafe fn try_from_raw_parts(base: *const u8, len: usize) -> Result<Self, IoVecError> {
+        if len > IOSIZE_MAX {
+            return Err(IoVecError::MaxLenExceeded);
+        }
+
+        Ok(Self { base, len })
+    }
+
+    /// This is for when the PS4 DOES NOT check the length (such as in recvmsg, recvfrom, sendmsg and sendto)
```
mikusp, line 21: Those comments are confusing, e.g. does recvmsg check the length or not? (or both?)

mikusp (dismissed):

src/kernel/src/fs/uio.rs
SuchAFuriousDeath: I copied it and forgot to change it, thanks

mikusp (approved):

PR #721, opened 2024-03-02
Initializes sockgetopt and socksetopt

SuchAFuriousDeath:


src/kernel/src/net/mod.rs
```diff
@@ -217,5 +288,25 @@ impl Display for AddressFamily {
     }
 }
 
+enum SockOpt {}
+
+#[derive(Debug, Error, Errno)]
+enum SetOptError {
+    #[error("Invalid value or value length")]
```
ultimaweapon, line 295: The message should follow Rust guideline.

src/kernel/src/net/mod.rs
```diff
@@ -217,5 +288,25 @@ impl Display for AddressFamily {
...
+    #[errno(EFAULT)]
+    InvalidValue,
+
+    #[error("Invalid length")]
```
ultimaweapon, line 299: Same here.

src/kernel/src/net/mod.rs
```diff
@@ -161,6 +188,52 @@ impl NetManager {
...
+    fn setsockopt(
+        &self,
+        fd: i32,
+        level: i32,
+        name: i32,
+        value: *const u8,
+        len: i32,
+        td: &VThread,
+    ) -> Result<(), SetOptError> {
+        if value.is_null() || len == 0 {
```
ultimaweapon, line 202: I don't think the condition here is correct.

src/kernel/src/net/socket.rs
```diff
@@ -23,7 +26,7 @@ impl Socket {
     pub fn new(
         domain: i32,
         ty: i32,
-        proto: i32,
+        proto: Option<NonZeroI32>,
```
ultimaweapon, line 29: Why this one is non-zero? What I saw in the `socreate` is it have a path for zero proto.

src/kernel/src/net/mod.rs
```diff
@@ -217,5 +288,25 @@ impl Display for AddressFamily {
...
+
+    #[error("Invalid length")]
+    #[errno(EINVAL)]
+    InvalidLength,
+}
+
+#[derive(Debug, Error, Errno)]
+enum GetOptError {
+    #[error("invalid length")]
+    #[errno(EFAULT)]
```
ultimaweapon, line 307: This seems incorrect.

ultimaweapon (changes requested):

src/kernel/src/net/socket.rs
SuchAFuriousDeath: It's  not NonZero, it's Option<NonZero>. The reason I did this is because It looks like 0 is a special-case value that's treated differently. Sound kind of like an option to me.
image
image

ultimaweapon: Then this looking good.

ultimaweapon (approved):

PR #723, opened 2024-03-03
Fixes incorrect mount path for game

ultimaweapon:


src/kernel/src/main.rs
```diff
@@ -215,18 +214,31 @@ fn start() -> Result<(), KernelError> {
         return Err(KernelError::MountFailed(path.into(), e));
     }
 
-    let path = VPathBuf::try_from(format!(
-        "/mnt/sandbox/{}_000/app0",
-        args.game.file_name().unwrap().to_str().unwrap()
-    ))
-    .unwrap();
+    // TODO: Check permission of these paths on the PS4.
+    let paths = [vpath!("/mnt/sandbox"), vpath!("/mnt/sandbox/pfsmnt")];
```
SuchAFuriousDeath, line 218: I originally mkdir'd these paths too, but I though that they always exist inside tmpfs

SuchAFuriousDeath (approved):

SuchAFuriousDeath:
And then we just change the root of the process to /mnt/sandbox/pfsmnt/{}-app0, right? I wasn't so far off :sweat_smile:

src/kernel/src/main.rs
ultimaweapon: All files in the tmpfs will live in the memory.

ultimaweapon:
> And then we just change the root of the process to /mnt/sandbox/pfsmnt/{}-app0, right? I wasn't so far off 😅

It should be a nullfs that mount this path.

ultimaweapon:
Seems like there is a bug in Actions because it should add `S-ready`.

SuchAFuriousDeath:
Github battle part 2 incoming? :D

SuchAFuriousDeath:
By the way, I would've implemented tempfs lookup already, but I'm not sure I know how. It seems like we need to get the actual type from dyn VnodeBackend

SuchAFuriousDeath:
Wait what? That makes no sense. It's actually fine, I just got confused yesterday

PR #725, opened 2024-03-03
Mounts /app0 and /dev on game root

ultimaweapon:
This also change the root of the process to the correct value.


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Looks like we could use mkdirall :D

PR #726, opened 2024-03-03
Updates README

ultimaweapon:
To clarify why we take so long on this while the other emulators already able to boot some games.


SuchAFuriousDeath (approved):

ultimaweapon:
Okay it is actually a bug in Actions somehow. No idea why compare commit did not return behind in this case.

PR #727, opened 2024-03-03
Implements sys_socketclose

SuchAFuriousDeath:


ultimaweapon (approved):
I wonder why Sony need this one instead of using a normal `close`.

SuchAFuriousDeath:
Maybe they needed it in a lower version of the firmware, or they made it 'just in case' they needed some extra logic for closing sockets. Who knows

PR #729, opened 2024-03-03
Refactors character devices

SuchAFuriousDeath:


SuchAFuriousDeath:
I want to know your opinion on this. By the way, I mentioned under some issue or PR that I checked a lot of devices and none of the seem to actually us FdOpen anywhere. We might be able to Rustify the devices.
Keep in mind, this is just a show off of what I have in mind.

SuchAFuriousDeath:
With this approach, driver-specific fields that FreeBSD stores in cdev->si_drv1 (I think) would be store inside the type that implements the DeviceTrait and any global data for all instances of a device would be stored probably in a static OnceLock.

SuchAFuriousDeath:
Hmmm. Besides, we might be able to do fdopen in a single function, probably, right? Just pass a bool to it, or an enum, or something like that

ultimaweapon:
If you don't want your idea to be missing better to start it as a discussion.

ultimaweapon:
> Hmmm. Besides, we might be able to do fdopen in a single function, probably, right? Just pass a bool to it, or an enum, or something like that

My idea is we should find an alternative solution where the PS4 call into this function. If we can fix at the call site then we can remove it.

SuchAFuriousDeath:
> > @ultimaweapon I want to know your opinion on this. Keep in mind, this is just a show off of what I have in mind. I mentioned under some issue or PR that I checked a lot of devices and none of the seem to actually us FdOpen anywhere. We might be able to Rustify the devices.
> 
> IIRC there is one place that call into this (can't remember where is this). It is the reason why I bring it into our code. When I was implemented this I want to remove it too due to it is so awkward.

So... what now? Will you merge this with the idea that we'll refactor the devices later, once we start encountering them? 
It's kind of difficult to predict what will happen if we're not reaching any :D

SuchAFuriousDeath:
Besides, didn't we jump ahead a bit too much with devfs? Maybe we should have stubbed it. Can it even happen that the file in VnodeBackend::open is null for us?

ultimaweapon:
> > > @ultimaweapon I want to know your opinion on this. Keep in mind, this is just a show off of what I have in mind. I mentioned under some issue or PR that I checked a lot of devices and none of the seem to actually us FdOpen anywhere. We might be able to Rustify the devices.
> > 
> > 
> > IIRC there is one place that call into this (can't remember where is this). It is the reason why I bring it into our code. When I was implemented this I want to remove it too due to it is so awkward.
> 
> So... what now? Will you merge this with the idea that we'll refactor the devices later, once we start encountering them? It's kind of difficult to predict what will happen if we're not reaching any :D

I don't think putting all devices in a single place is a good idea. The `fs/dev` should be a foundation for creating a devices instead of implementing it own devices.

> Besides, didn't we jump ahead a bit too much with devfs? Maybe we should have stubbed it and done it later. Can it even happen that the file in VnodeBackend::open is null for us?

Do you mean the devfs system or the device files?

SuchAFuriousDeath:
The filesystem and vnodeops

ultimaweapon:
> And also, I'm not sure if you're talking about fdopen being called in VnodeBackend::open, or you've seen it in an actual PS4 device. Like I said, in FreeBSD there's only one driver that actually uses it. I looked at a lot of PS4 devices and none of the use it too :D

Better to continue this at #730.

ultimaweapon:
> The filesystem and vnodeops

I think it is the right decision that we decided to rework on FS system. This allow us to implement anything that rely on FS correctly from the beginning without the need to come back later to rewrite it in the future.

If we keep going without fixing the root problem anything that depend on it will also goes wrong, which will cause the problem to accumulate and it will be much harder and harder to fix. Once we have too much problems instead of too many things that not yet implemented we will be exhausted because we don't want to fix the problem.

SuchAFuriousDeath:
Because now, there's and open issue for a 'problem', when it feels like the real problem is that fdopen shouldn't be in our code in the first place.

SuchAFuriousDeath:
> I don't think putting all devices in a single place is a good idea. The `fs/dev` should be a foundation for creating a devices instead of implementing it own devices.

Really? For example. FreeBSD puts a lot of them in one directory. Where would you put them?

SuchAFuriousDeath:
Source: http://fxr.watson.org/fxr/source/dev/?v=FREEBSD-9-STABLE

ultimaweapon:
> > I don't think putting all devices in a single place is a good idea. The `fs/dev` should be a foundation for creating a devices instead of implementing it own devices.
> 
> Really? For example. FreeBSD puts a lot of them in one directory. Where would you put them?

Where did you see that? [Here](https://github.com/freebsd/freebsd-src/tree/release/9.1.0/sys/fs/devfs) is what I saw.

ultimaweapon:
> Source: http://fxr.watson.org/fxr/source/dev/?v=FREEBSD-9-STABLE

It seems like those for hardware supports, not the device file.

SuchAFuriousDeath:
To be clear, I'm not saying that all the devices should be in fs/dev, I'm saying that we should keep them together.

> Where did you see that? [Here](https://github.com/freebsd/freebsd-src/tree/release/9.1.0/sys/fs/devfs) is what I saw.

Here is what I saw: http://fxr.watson.org/fxr/source/dev/?v=FREEBSD-9-STABLE

SuchAFuriousDeath:
> > Source: http://fxr.watson.org/fxr/source/dev/?v=FREEBSD-9-STABLE
> 
> It seems like those for hardware supports, not the device file.

They contain the kernel modules that make the devices, though:
http://fxr.watson.org/fxr/ident?v=FREEBSD-9-STABLE&i=cdevsw

ultimaweapon:
We can try putting it in a directory on the top-level (not in the `fs`).

SuchAFuriousDeath:
I moved them to dev in the kernel crate. I also removed the fdopen. This is enough changes for now I think.

SuchAFuriousDeath:
Apparently, my brilliant idea isn't object safe :(

src/kernel/src/dev/mod.rs
ultimaweapon, line 1: All devices should have a unique name so better to re-export all sub-modules here.

ultimaweapon (changes requested):

src/kernel/src/dev/dmem.rs
```diff
@@ -0,0 +1,16 @@
...
+};
+use std::sync::Arc;
+
+#[derive(Debug)]
+struct Dmem {}
+
+impl Device for Dmem {
+    #[allow(unused_variables)] // TODO: remove when implementing
+    fn ioctl(self: Arc<Self>, cmd: IoCmd, td: Option<&VThread>) -> Result<(), Box<dyn Errno>> {
+        todo!()
```
ultimaweapon, line 14: Why not bringing what we have implemented in the `src/kernel/src/fs/dev/dmem1.rs` here?

ultimaweapon (changes requested):

src/kernel/src/dev/dmem.rs
SuchAFuriousDeath: Good point

SuchAFuriousDeath:
@ultimaweapon By the way, VThread holds a strong reference to VProc and VProc holds a vec of strong references to VThreads, doesn't that leak memory?

ultimaweapon:
> @ultimaweapon By the way, VThread holds a strong reference to VProc and VProc holds a vec of strong references to VThreads, doesn't that leak memory?

Nope. There is a RAII object that will remove it from the process when the thread is terminated.

src/kernel/src/fs/ioctl.rs
```diff
@@ -96,9 +118,27 @@ commands! {
...
+type Unknown2 = Unknown<2>;
+type Unknown8 = Unknown<8>;
+type Unknown16 = Unknown<16>;
+
+/// A dummy type to be used as a placeholder for unknown data.
+pub struct Unknown<const N: usize>([u8; N]);
+
+impl<const N: usize> Debug for Unknown<N> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Unknown type with size of {N}")
```
ultimaweapon, line 142: What wrong with derive macro?

ultimaweapon (changes requested):

src/kernel/src/dev/dipsw.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::sync::Arc;
+use thiserror::Error;
+
+#[derive(Debug)]
+struct Dipsw {}
+
+impl Device for Dipsw {
+    fn ioctl(self: Arc<Self>, cmd: IoCmd, td: &VThread) -> Result<(), Box<dyn Errno>> {
+        if !td.cred().is_system() {
+            return Err(Box::new(IoctlErr::InsufficientCredentials));
```
ultimaweapon, line 16: What I saw is there are some commands that can be execute without a system credential.

ultimaweapon (changes requested):

src/kernel/src/dev/dipsw.rs
SuchAFuriousDeath: Soooo.... Option\<&VThread\>?

ultimaweapon: I mean there are some commands that can be execute if `sceSblACMgrIsSystemUcred` return 0. So return error here is not correct.

SuchAFuriousDeath: Oh :eyes:

SuchAFuriousDeath: You're right. There's so many nested ifs that I overlooked that. Fixed

SuchAFuriousDeath:
@ultimaweapon I think that to do Open, we will need Box<dyn FnOnce() -> Box<dyn Device>> or something like that.

SuchAFuriousDeath:
@ultimaweapon I think I figured it out, take a look when you can.

src/kernel/src/fs/dev/cdev.rs
```diff
@@ -37,7 +37,7 @@ pub struct CharacterDevice {
 impl CharacterDevice {
     /// See `devfs_alloc` on the PS4 for a reference.
     pub(super) fn new(
-        sw: &Arc<CdevSw>,
+        driver: impl Device,
```
ultimaweapon, line 40: This one should be a last parameter so we can write:

```rust
CharacterDevice::new(..., SomeDevice {
    a: b,
    c: d,
    ...
});
```

src/kernel/src/fs/dev/cdev.rs
```diff
@@ -68,8 +68,8 @@ impl CharacterDevice {
         }
     }
 
-    pub fn sw(&self) -> &CdevSw {
-        self.sw.as_ref()
+    pub fn driver(&self) -> &dyn Device {
+        self.driver.as_ref()
```
ultimaweapon, line 72: The `driver` should not exposed to the outside otherwise its methods can be called with incorrect `CharacterDevice`.

src/kernel/src/fs/dev/cdev.rs
```diff
@@ -166,28 +166,6 @@ bitflags! {
     }
 }
 
-/// An implementation of `cdevsw` structure.
-#[derive(Debug)]
-pub struct CdevSw {
-    flags: DriverFlags, // d_flags
-    open: CdevOpen,     // d_open
```
ultimaweapon, line 173: Any reason why not remove `CdevOpen`?

src/kernel/src/fs/dev/cdev.rs
```diff
@@ -237,5 +233,5 @@ pub(super) enum DefaultError {
 
     #[error("ioctl not supported")]
     #[errno(ENODEV)]
-    IoctlNotSupported,
+    CommandNotSupported,
```
ultimaweapon, line 236: `IoctlNotSupported` looks better than the new version in this case because `IoctlNotSupported` means the method is not supported regardless the `cmd` passed in.

ultimaweapon (changes requested):

src/kernel/src/fs/dev/cdev.rs
SuchAFuriousDeath: Blindness :D

src/kernel/src/tty/mod.rs
```diff
@@ -17,40 +14,25 @@ pub struct TtyManager {
 impl TtyManager {
```
ultimaweapon, line 14: This one should be moved to `dev` too.

ultimaweapon (changes requested):

src/kernel/src/tty/mod.rs
SuchAFuriousDeath: Where exactly? Inside ttyconsdev.rs?

ultimaweapon:
@SuchAFuriousDeath do you know if we are going to need `d_flags` in the `cdevsw`? I'm worried about removing it right now because I'm not sure if we are going to need it in the future. If you are unsure better to keep it in each implementation for now.

SuchAFuriousDeath:
Good point, I didn't even realize

SuchAFuriousDeath:
@ultimaweapon Take a look

ultimaweapon (changes requested):

src/kernel/src/dev/ttyconsdev.rs
```diff
@@ -0,0 +1,93 @@
...
+    make_dev, CharacterDevice, DeviceDriver, DriverFlags, IoCmd, MakeDevError, MakeDevFlags, Mode,
+    OpenFlags, Uio, UioMut,
+};
+use crate::ucred::{Gid, Uid};
+use crate::{errno::Errno, process::VThread};
+use std::sync::Arc;
+use thiserror::Error;
+
+#[derive(Debug)]
+pub struct TtyConsDev {}
```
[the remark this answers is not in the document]
SuchAFuriousDeath: Well, it's named that way in FreeBSD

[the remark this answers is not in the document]
SuchAFuriousDeath: image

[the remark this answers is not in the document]
ultimaweapon: We don't need to follow how FreeBSD does if we have a better reason to not follow.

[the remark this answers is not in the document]
SuchAFuriousDeath: I agree. Commits are up

src/kernel/src/debug/mod.rs
ultimaweapon, line 1: `DebugManager` should be moved to the same module as `Deci` the same as TTY console.

ultimaweapon (changes requested):

src/kernel/src/debug/mod.rs
SuchAFuriousDeath: Done

src/kernel/src/dev/deci.rs
ultimaweapon, line 1: `Deci` and `DeciDev` is very confusing name. Maybe rename `Deci` to `Driver` and make it private then rename `DeciDev` to juse `Deci`?

ultimaweapon (changes requested):

src/kernel/src/dev/deci.rs
SuchAFuriousDeath: done

src/kernel/src/dev/random.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    fn ioctl(
+        &self,
+        dev: &Arc<CharacterDevice>,
+        cmd: IoCmd,
+        _: &VThread,
+    ) -> Result<(), Box<dyn Errno>> {
+        match cmd {
+            IoCmd::FIOASYNC(_) | IoCmd::FIONBIO(_) => Ok(()),
+            _ => Err(Box::new(DefaultDeviceError::IoctlNotSupported)),
```
ultimaweapon, line 40: I don't think this error is correct.

ultimaweapon (changes requested):

src/kernel/src/dev/random.rs
SuchAFuriousDeath: FIxed

ultimaweapon (approved):

PR #732, opened 2024-03-03
Implements poll events

SuchAFuriousDeath:


SuchAFuriousDeath:
The flags in PollEvents are the same as FreeBSD, this can be deduced through inspecting poll_no_poll. I also think PollFd makes more sense to have in filedesc, because that's where pollscan will be in the future alongside other functions related to polling.

src/kernel/src/fs/file.rs
```diff
@@ -226,6 +226,9 @@ pub trait FileBackend: Debug + Send + Sync + 'static {
         Err(Box::new(DefaultError::IoctlNotSupported))
     }
 
+    #[allow(unused_variables)]
+    fn poll(self: &Arc<Self>, file: &VFile, events: PollEvents, td: &VThread) -> PollEvents;
+
```
ultimaweapon, line 231: Can we put the name on FreeBSD in its documentation so it is easier to follow the code?

src/kernel/src/process/filedesc.rs
```diff
@@ -151,6 +150,36 @@ impl FileDesc {
...
+pub struct PollFd {
+    fd: i32,
+    events: PollEvents,
+    revents: PollEvents,
+}
+
+bitflags! {
+    #[repr(transparent)]
+    #[derive(Debug, Clone, Copy)]
+    pub struct PollEvents: u16 {
```
ultimaweapon, line 163: This one on the PS4 is `u16`?

ultimaweapon (changes requested):

src/kernel/src/process/filedesc.rs
SuchAFuriousDeath: in PollFd it's an i16 (so u16 is acceptable I think), but then the poll functions in cdevsw and fileops return an int just like all the other functions. This returned int(i32 for us) is then set into the i16 of the pollfd struct.

src/kernel/src/fs/file.rs
SuchAFuriousDeath: What do you mean? Like 'fo_poll'?

ultimaweapon: Something like:

```rust
/// An implementation of `fo_poll`.
```

src/kernel/src/process/filedesc.rs
ultimaweapon: If its size is inconsistent better to not use `repr(transparent)` here and put a warning in its documentation so people won't accept or pass this type directly between the PS4.

SuchAFuriousDeath: There is no problem here. The only place this is received from the PS4 is in sys_poll and we are matching that size. All the other functions are internal and we just return PollEvents instead of and int. The higher bits would've been discarded anyway

ultimaweapon: The problem here is this type used with `FileBackend::poll`, which represents an implementation of `fo_poll`. Once I saw this I will immediately assume this type has the same representation as the PS4 version because it has `#[repr(transparent)]`.

SuchAFuriousDeath: Is it enough to just add a comment or you want something more?

ultimaweapon: We need to remove `#[repr(transparent)]` and document this type that it cannot be passed directly from/to the PS4.

SuchAFuriousDeath: But it can be passed, because it's the right size

ultimaweapon: The point here is its size inconsistent. Yes it can be used directly in some functions. But some functions this type have different sizes.

SuchAFuriousDeath: But again, why does it matter? All those functions that return a bigger size are purely internal and we can just return PollEvents and the behaviour will be the same. If all the information can always fit in a u16, why bother? We can just add a comment explaining our choice and then nothing bad can ever happen

SuchAFuriousDeath: That was my intention

src/kernel/src/process/filedesc.rs
ultimaweapon, line 1: I don't think this is a correct place for `PollFd` and `PollEvents`. It should be in the `fs` instead.

ultimaweapon (changes requested):

src/kernel/src/process/filedesc.rs
SuchAFuriousDeath: I think it might be the correct place. Polling isn't necesarilly tied to the filesystem, but it is inherently tied to file descriptors. I actually went a little further then what is in this PR and FileDesc will probably have the pollscan method. But I don't really care where it is, it's your choice.

ultimaweapon: The reason it should be in the `fs` is because it is a place for FS related syscall handlers. We have everything that deal with file descriptor there and then forward it do the `filedesc` in the process.

ultimaweapon (approved):

Issue #735, opened 2024-03-05
Make create_thread unsafe


ultimaweapon:
https://github.com/obhq/obliteration/blob/main/src%2Fkernel%2Fsrc%2Fprocess%2Fmod.rs#L460

ultimaweapon:
Wait, this is already correct. But the create thread method that accepts pointers should be unsafe.

PR #736, opened 2024-03-05
Makes create_thread unsafe

SuchAFuriousDeath:
#735 
Perhaps thr_new should be unsafe as well?


ultimaweapon:
> #735 
> Perhaps thr_new should be unsafe as well? 

Good idea.

mikusp (approved):

PR #747, opened 2024-03-18
Implements Read and Seek on VFile, removes fdopen

SuchAFuriousDeath:
This is more of a rough sketch. I had the VnodeBackend::open in mind a long time ago. I think that if you look at implementations of vop_open, the file* is basically passed to them just so that they can call finit on it, which is the equivalent of returning the filetype and setting it in open as I did.

This PR
- makes VnodeBackend::open return a VFileType (I thinks that makes sense, since that is the reason of passing a file * in the first place. I'm not sure how it can happen that the file is null, if you don't know either, I suggest we cross that bridge when we get there.
- Implements Read and partially implements Seek
- Implements exfatfs open:
- Mounts another nullfs such that it mirrors /system/common/lib (So that we can look up the libraries).
- Adds an offset: i64 argument to reads and writes. Perhaps this should be inside Uio and UioMut, but I don't know if it has to.
- That's all, folks

I'll be gradually splitting this into multiple PRs so that it's easier to review.

Things to consider: 
Is it ok to do pwritev every time in HostFile? It's probably possible to store the next expected offset and if it matches the provided one, do a simple writev.


SuchAFuriousDeath:
@ultimaweapon What do you think about using OwnedFd in HostFile?

SuchAFuriousDeath:
@ultimaweapon FYI, we currently fail on looking up libkernel
image

SuchAFuriousDeath:
I'm past that, now the read seems to fail

SuchAFuriousDeath:
Ok, I got the path wrong, so I was trying to read a directory, oops. Now we're failing at devfs getattr which is being called be revoke :heart_eyes:

SuchAFuriousDeath:
We're currently stuck at opening dipsw, which isn't implemented for us

SuchAFuriousDeath:
Closing this as it has been split into several smaller PRs

PR #749, opened 2024-03-19
Fixes mutex issues in sys_dynlib_load_prx, removes VnodeBackend::open

SuchAFuriousDeath:
Calling load causes a panic in sys_dynlib_load_prx, because we're still holding onto the gutex guard


src/kernel/src/fs/vnode.rs
```diff
@@ -304,12 +303,9 @@ pub(super) trait VnodeBackend: Debug + Send + Sync + 'static {
...
         &self,
         #[allow(unused_variables)] vn: &Arc<Vnode>,
-        #[allow(unused_variables)] td: Option<&VThread>,
         #[allow(unused_variables)] mode: OpenFlags,
-        #[allow(unused_variables)] file: Option<&mut VFile>,
-    ) -> Result<(), Box<dyn Errno>> {
-        Ok(())
-    }
+        #[allow(unused_variables)] td: Option<&VThread>,
+    ) -> Result<VFileType, Box<dyn Errno>>;
```
ultimaweapon, line 308: Does not this suppose to return `Option<VFileType>`?

ultimaweapon (changes requested):

src/kernel/src/fs/vnode.rs
SuchAFuriousDeath: Why should it? If you want to represent the deadfileops case, we can just put what would happen directly inside the VnodeBackend implementation
image

SuchAFuriousDeath:
@ultimaweapon By the way, how do you plan to create all the other devices? Will they all have a separate manager?

src/kernel/src/fs/vnode.rs
ultimaweapon: So this is the only cases for null file?

ultimaweapon:
> @ultimaweapon By the way, how do you plan to create all the other devices? Will they all have a separate manager?

Unless we have a better idea.

src/kernel/src/fs/vnode.rs
ultimaweapon: IIRC some of FS does not implement `vop_open`, thus leaving `a_fp` to be null on return. How we will handled this case?

SuchAFuriousDeath: hmmmm

SuchAFuriousDeath: I'm confused, to be honest. Isn't this a separate issue?

ultimaweapon: Nope because you change how our implementation of `vop_open` works. On PS4 most call to `vop_open` pass `file` as a null and most of FS does not even update it on return.

SuchAFuriousDeath: I understand now

SuchAFuriousDeath: We could add a method like open_nofile to VnodeBackend for that once we need it.

SuchAFuriousDeath: Wait. Is that even a problem? We change just use open and then not use the return VFileType, no? Or what exactly is the issue?

ultimaweapon: What `VFileType` we should return from the FS that does no update the input `file`? The correct solution here should be change the type of `file` parameter to `Option<&mut VFileType>` instead of return it from the method.

SuchAFuriousDeath: But that allows us to set all fields that have a setter and if we have a setter method for flags and ty, they can then be used elsewhere. I think we should just make a separate method for the fp = null case when we reach it.

ultimaweapon: Your answer does not solve my first paragraph.

SuchAFuriousDeath: You're right. Which are the filesystems that act in this way? Even if we use Option\<&mut VFile>, how do we deal with the fact that the filesystems you're talking about don't mutate the file* at all? Couldn't the open function of the filesystems be (theoretically) called from places where FreeBSD expects finit to be called on the file*?
If this happened in FreeBSD, I suppose we should keep my remake of what FreeBSD does and where the filesystem doesn't mutate the file *, we just return VFileType::Vnode, which is what would happen when FreeBSD passes a non-null file * to VOP_OPEN on a vnode from a filesystem that doesn't mutate it.

SuchAFuriousDeath: Hopefully, I'm not missing something.
I think the function should probably return (VFileFlags, VFileType) actually, but that's just a detail, I'll change it in this PR when I have time.

ultimaweapon: I think we should understand the purpose of `fp` before processing with refactor.

SuchAFuriousDeath: image

ultimaweapon: We need specific meaning of `such information`.

SuchAFuriousDeath: So what can I do?

SuchAFuriousDeath: I don't really know what reverting gives us, I already pointed out that the issues with Option<&mut VFile> are basically the same as with this refactor.

SuchAFuriousDeath: e.g. if a Filesystem doesn't mutate the file*, and you're in Fs::open, what VFileType will the file have?

ultimaweapon: Actually I don't really understand the point of `vop_open`. It does not make any senses to me. Do you know how this system is working?

SuchAFuriousDeath: I suppose that it¨s initialization logic when opening a file (Altough, not necesarilly, since the pointer can be null). Obviously, if any filesystem changes anything else other than  the file flags or the type and data, the logic in this PR would have to be refactored.
TLDR: I don't fully understand it either

ultimaweapon: Maybe we should try to understand this first. We might eventually don't need it, which will solve `file` issue automatically.

ultimaweapon: Okay I got an idea. Just remove the whole `vop_open` for now. In the future when we understand how it work we can bring it back.

SuchAFuriousDeath: How do se set the vfile type then? It depends on the filesystem?

ultimaweapon: I don't think that value come from the filesystem driver. It should come from the higher level. AFAIK the primary concern of the filesystem driver is `vnode`, not `file`.

SuchAFuriousDeath: Uhhhhmmm ok. So how exactly do we find out which VFileType to set?

ultimaweapon: That should be the job of the function that construct the `VFile`.

SuchAFuriousDeath: Hmmm. ok

SuchAFuriousDeath: Take a look. Altough, I'm not quite sure what to do in the open function, FreeBSD runs an insane amount of code.

src/kernel/src/fs/mod.rs
```diff
@@ -1113,7 +1117,10 @@ pub enum MountError {
 #[derive(Debug, Error, Errno)]
 pub enum OpenError {
     #[error("cannot lookup the file")]
-    LookupFailed(#[source] LookupError),
+    LookupFailed(#[from] LookupError),
```
ultimaweapon, line 1120: https://github.com/obhq/obliteration/discussions/768

src/kernel/src/fs/mod.rs
```diff
@@ -1113,7 +1117,10 @@ pub enum MountError {
...
+
+    #[error("cannot open the file")]
+    OpenFailed(#[source] Box<dyn Errno>),
```
ultimaweapon, line 1123: This should not needed anymore.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #754, opened 2024-03-22
Add path specification for Windows Root

VocalFan:
The simplest solution I could think of is solving this internally instead of users having to reconfigure their registry or .kernel-debug file.

Fixes the Root function for windows by using the `\??\` specification for the drive path. `\\??\\` due to how Rust handles backspaces.

We now reach the todo for raw_mkdir for Windows.

### DIFF
--- a/src/kernel/src/fs/host/file.rs
+++ b/src/kernel/src/fs/host/file.rs
@@ -49,9 +49,8 @@ impl HostFile {
         use windows_sys::Win32::System::Kernel::OBJ_CASE_INSENSITIVE;
 
         // Encode path name.
-        let mut path_spec = String::from("\\??\\");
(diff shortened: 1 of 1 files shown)


src/kernel/src/fs/host/file.rs
```diff
@@ -49,7 +49,9 @@ impl HostFile {
         use windows_sys::Win32::System::Kernel::OBJ_CASE_INSENSITIVE;
 
         // Encode path name.
-        let path = path.as_ref();
+        let mut path_spec = String::from("\\??\\");
```
SuchAFuriousDeath, line 52: I think this should be PathBuf::from, with the next lines adjusted accordingly.

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath (approved):

PR #759, opened 2024-03-23
Maps VM memory

ultimaweapon:
The reason we still use C++ with KVM is because https://crates.io/crates/kvm-ioctls is too high-level. We need something like raw KVM binding.

### DIFF
--- a/src/hv/build.rs
+++ b/src/hv/build.rs
@@ -1,9 +1,13 @@
 fn main() {
     match std::env::var("CARGO_CFG_TARGET_OS").unwrap().as_str() {
-        "linux" | "android" => cc::Build::new()
-            .cpp(true)
-            .file("src/linux/kvm.cpp")
-            .compile("hvkvm"),
+        "linux" | "android" => {
+            println!("cargo::rerun-if-changed=src/linux/kvm.cpp");
+
+            cc::Build::new()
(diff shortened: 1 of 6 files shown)


SuchAFuriousDeath:
Found this: https://crates.io/crates/kvm-ioctls

ultimaweapon:
> Found this: https://gitlab.com/preon/libkvm

Last commit was 5 years ago 😱

SuchAFuriousDeath:
Maybe we could make our own binding crate

SuchAFuriousDeath:
> > Found this: https://gitlab.com/preon/libkvm
> 
> Last commit was 5 years ago 😱

That's not necesarilly bad though, is it? Maybe there's just nothing to do because the api doesn't change.

ultimaweapon:
> Maybe we could make our own binding crate

I don't think it wort the effort than using `cc` to build the C++ wrapper.

> > > Found this: https://gitlab.com/preon/libkvm
> > 
> > 
> > Last commit was 5 years ago 😱
> 
> That's not necesarilly bad though, is it? Maybe there's just nothing to do because the api doesn't change.

Just checking its code and it is also high-level the same as the first one.

SuchAFuriousDeath (approved):

PR #767, opened 2024-03-25
Implement evf syscalls

mikusp:


src/kernel/src/idt/mod.rs
```diff
@@ -68,6 +68,23 @@ impl<T> Idt<T> {
         Ok((value, id))
     }
 
+    /// See `id_rlock` on the PS4 for a reference.
+    pub fn get(&self, id: usize, ty: Option<u16>) -> Option<&Entry<T>> {
```
SuchAFuriousDeath, line 72: Taking an Option\<u16\> Seems weird to me. I think we should take a u16 and then below this, do 
let set  = self.sets.get(i / 0x80)?;

src/kernel/src/evf/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        let num_waiting_threads: *mut i32 = i.args[2].into();
+
+        info!(
+            "sys_evf_cancel({}, {:#x}, {:#x})",
+            id,
+            pattern,
+            (num_waiting_threads as usize)
+        );
+
+        match self.proc.objects_mut().get(id, Some(EventFlag::ENTRY_TYPE)) {
```
SuchAFuriousDeath, line 278: This should be 

let entry = self.proc.objects_mut().get(id, Some(EventFlag::ENTRY_TYPE)).ok_or(SysErr::Raw(ESRCH))?;

That gets rid of the match statement

src/kernel/src/evf/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+    }
+
+    fn sys_evf_clear(self: &Arc<Self>, _: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
+        let id = i.args[0].into();
+        let pattern: u64 = i.args[1].into();
+
+        info!("sys_evf_clear({}, {:#x})", id, pattern);
+
+        match self.proc.objects_mut().get(id, Some(EventFlag::ENTRY_TYPE)) {
```
SuchAFuriousDeath, line 248: same here

src/kernel/src/evf/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        todo!()
+    }
+
+    fn sys_evf_set(self: &Arc<Self>, _: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
+        let evf = i.args[0].into();
+        let pattern: u64 = i.args[1].into();
+
+        info!("sys_evf_set({}, {:#x})", evf, pattern);
+
+        match self
```
SuchAFuriousDeath, line 182: same here

src/kernel/src/evf/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+            return Err(SysErr::Raw(EINVAL));
+        }
+
+        if !timeout.is_null() {
+            todo!()
+        }
+
+        let objects = self.proc.objects();
+
+        match objects.get(id, Some(EventFlag::ENTRY_TYPE)) {
```
SuchAFuriousDeath, line 114: Same here

SuchAFuriousDeath (changes requested):

PR #770, opened 2024-03-26
sys_batch_map

mikusp:


src/kernel/src/process/vm/mod.rs
```diff
@@ -838,3 +969,43 @@ pub enum MemoryUpdateError {
...
 }
+
+#[repr(C)]
+struct BatchMapOp {
+    addr: usize,
+    offset: usize,
+    len: usize,
+    prot: u8,
+    ty: u8,
+    op: MapOp,
```
SuchAFuriousDeath, line 980: I think using an enum here causes potentially undefined behavirour when you create a *BatchMapOp from a SysArg.

SuchAFuriousDeath (changes requested):

src/kernel/src/process/vm/mod.rs
mikusp: Is using i32 better?

SuchAFuriousDeath: It depends on how many bytes the PS4 reads. If 4, then yes.

mikusp: Seems to read 4, I wrote it a bit different

src/kernel/src/process/vm/mod.rs
```diff
@@ -594,6 +605,115 @@ impl MemoryManager {
...
+
+        let slice_size = num_of_ops.try_into().ok().ok_or(SysErr::Raw(EINVAL))?;
+        let operations = unsafe { std::slice::from_raw_parts(operations, slice_size) };
+
+        let mut ret = Ok(0.into());
+        let mut processed = 0;
+        for op in operations {
+            // TODO: implement remaining logic when op != MapTypeProtect && some addr != 0
+
+            ret = match op.op.try_into() {
```
SuchAFuriousDeath, line 627: There should be a ? here I think, then You don't have to wrap every variant in Ok(..)

SuchAFuriousDeath (changes requested):

src/kernel/src/process/vm/mod.rs
mikusp: I think that'll behave differently, returning Err without breaking the loop and writing the number of processed entries

SuchAFuriousDeath: Yeah, you're right. Oops. We could really use try_blocks here, but they're not stable :(

src/kernel/src/vm/mod.rs
```diff
@@ -582,6 +594,115 @@ impl Vm {
...
+        if flags.bits() & 0xe0bffb6f != 0 {
+            return Err(SysErr::Raw(EINVAL));
+        }
+
+        let slice_size = num_of_ops.try_into().ok().ok_or(SysErr::Raw(EINVAL))?;
+        let operations = unsafe { std::slice::from_raw_parts(operations, slice_size) };
+
+        let mut ret = Ok(0.into());
+        let mut processed = 0;
+        for op in operations {
```
SuchAFuriousDeath, line 613: I would do for batch_op in operations, so that you don't have op.op below

SuchAFuriousDeath (changes requested):

src/kernel/src/vm/mod.rs
SuchAFuriousDeath, line 613: You might be able to use operations.iter().try_for_each() I think. The ? operator could then be used inside the closure that's passed

SuchAFuriousDeath (changes requested):

mikusp:
I've fixed what was previously requested, can we get back to the rest once dmem is somewhat working? 😂

SuchAFuriousDeath (approved):

PR #772, opened 2024-03-26
Fixes raw_open and microtime on windows

SuchAFuriousDeath:


VocalFan (approved):
LGTM! (Also nice fix for the microtime :P)

ultimaweapon:
@VocalFan have you understand what the code does before approving? As I said before that the rule of thumb is do not approve if you don't understand what the code does.

VocalFan:
Boy, I may be dumb in the deeper things... But we BOTH dived into this code before.

ultimaweapon:
If you understand then can you explain #773?

VocalFan:
> If you understand then can you explain #773?

Yep... File_OPEN only reads and opens the file if available, and errors if not available... Though normally, the /dev and /mnt folders already exist, you would... Want a fallback where if it DOES error, then create it with File_CREATE. Thing is it seems we have to do it in this order as File_CREATE just throws an error if the file does already exist.

VocalFan:
> If you understand then can you explain #773?

Hey, your implementation was broken too, so you can't be mad when we go from a non-working fix to a half-working fix that... If the folders already exist... We reach

```rs
    pub fn open(&self, path: impl AsRef<VPath>, td: Option<&VThread>) -> Result<VFile, OpenError> {
        let vnode = self
            .lookup(path, true, td)
            .map_err(OpenError::LookupFailed)?;

        todo!();
    }
```

ultimaweapon:
> > If you understand then can you explain #773?
> 
> Yep... File_OPEN only reads and opens the file if available, and errors if not available... Though normally, the /dev and /mnt folders already exist, you would... Want a fallback where if it DOES error, then create it with File_CREATE. Thing is it seems we have to do it in this order as File_CREATE just throws an error if the file does already exist.

https://github.com/obhq/obliteration/blob/main/src/kernel/src/fs/mod.rs#L111-L127

And `raw_mkdir` need to behave in the same way for both Windows and *nix, which is already correct before this PR. Next time please don't approve any PRs that has any changes on how the code work. You still can approve the other PRs like display string updates, grammar fix, etc.

PR #774, opened 2024-03-27
First try to read file, then create file if it doesn't exist.

### LINKED ISSUE #773: Fix raw_mkdir broken on Windows
Introduced by #772.

VocalFan:
Fix to SuchAFuriousDeath having the fix be half-right.

Is kind of unclean, but the best we have for now. Tested to handle folders existing and not existing just fine.

Fixes #773

### DIFF
--- a/src/kernel/src/fs/host/file.rs
+++ b/src/kernel/src/fs/host/file.rs
@@ -208,10 +208,12 @@ impl HostFile {
         use std::ptr::null_mut;
         use windows_sys::Wdk::{
             Foundation::OBJECT_ATTRIBUTES,
-            Storage::FileSystem::{NtCreateFile, FILE_DIRECTORY_FILE, FILE_OPEN},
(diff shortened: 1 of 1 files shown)


ultimaweapon:
This is not correct because it will cause behavior different from `raw_mkdir` on *nix. See https://github.com/obhq/obliteration/pull/772#issuecomment-2022144486 for more details.

VocalFan:
Ya say that, yet this still seems to give results :P

I mean, we reach the todo'd open function now... Instead of just erroring. Even then if the folders already existed, file_open wouldn't error and we'd reach Open anyways. But no, windows doesn't like having one type where it `if file exists, read it, if not, make then read it.` You have to do both.

SuchAFuriousDeath:
That's not true, 
image

VocalFan:
> That's not true, image

...Damn, I'll redo this then lmao-

SuchAFuriousDeath:
It will still be wrong

PR #776, opened 2024-03-27
define some ioctls

mikusp:


src/kernel/src/dmem/mod.rs
```diff
@@ -19,22 +21,34 @@ impl DmemManager {
...
-        let set: i32 = i.args[0].try_into().unwrap();
-        let get: i32 = td.proc().dmem_container().try_into().unwrap();
+        let dmem_id: i32 = i.args[0].try_into().unwrap();
 
-        if set != -1 {
-            todo!("sys_dmem_container with update != -1");
+        if dmem_id != -1 {
+            td.priv_check(Privilege::SCE685)?;
+
+            if dmem_id > 3 || dmem_id < -1 {
```
SuchAFuriousDeath, line 38: I think this check should be inside impl TryFom<i32> for DmemContainer

src/kernel/src/dmem/mod.rs
```diff
@@ -19,22 +21,34 @@ impl DmemManager {
...
+                //   todo: check if dmem_id device has not been created yet
+                //   || dmem_state[dmem_id].dev == None
+                return Err(SysErr::Raw(EINVAL));
+            }
+
+            *td.proc().dmem_container_mut() = dmem_id as usize;
```
SuchAFuriousDeath, line 44: This proc field would probably be of type DmemContainer then.

SuchAFuriousDeath (changes requested):

src/kernel/src/fs/ioctl.rs
```diff
@@ -105,6 +107,19 @@ commands! {
...
+        DMEMGETPRT(&mut PrtAperture) = 0xC018800C,
+        /// Get available memory size
+        DMEMGETAVAIL(&mut DmemAvailable) = 0xC0208016,
+        /// Allocate direct memory
+        DMEMALLOC(&mut DmemAllocate) = 0xC0288001,
+        /// Query direct memory
+        DMEMQUERY(&mut DmemQuery) = 0x80288012,
+
+        /// sceKernelMemoryPoolExpand
+        BPOOLEXPAND(&BlockpoolExpandArgs) = 0xC020A801,
```
SuchAFuriousDeath, line 120: Move the ones beginning with a B to the top, so that the variants are sorted 🥺

SuchAFuriousDeath (changes requested):

src/kernel/src/dmem/mod.rs
```diff
@@ -19,22 +21,32 @@ impl DmemManager {
...
-        if set != -1 {
-            todo!("sys_dmem_container with update != -1");
+        if dmem_id != -1 {
+            td.priv_check(Privilege::SCE685)?;
+
+            let dmem_container = dmem_id.try_into()?;
+            //   todo: check if dmem_id device has been created yet
+            //   || dmem_state[dmem_id].dev != None
+
+            *td.proc().dmem_container_mut() = dmem_container;
```
SuchAFuriousDeath, line 42: I think there should be a todo!() here for the moment being

src/kernel/src/dmem/mod.rs
```diff
@@ -19,22 +21,32 @@ impl DmemManager {
...
         }
 
-        Ok(get.into())
+        // todo: set td_retval[0] to old td.proc().dmem_container() value
+        Ok(0.into())
```
SuchAFuriousDeath, line 46: I think for us, returning is equivalent to setting td_retval[0]

SuchAFuriousDeath (changes requested):

src/kernel/src/dmem/mod.rs
SuchAFuriousDeath: So that means you should return get.into();

src/kernel/src/fs/ioctl.rs
```diff
@@ -85,6 +87,11 @@ macro_rules! commands {
 
 commands! {
     pub enum IoCmd {
+        /// sceKernelMemoryPoolExpand
+        BPOOLEXPAND(&BlockpoolExpandArgs) = 0xC020A801,
```
SuchAFuriousDeath, line 91: This one should be &mut

src/kernel/src/fs/ioctl.rs
```diff
@@ -105,6 +112,14 @@ commands! {
...
         /// Get total size?
         DMEM10(&mut usize) = 0x4008800a,
+        /// Get PRT aperture
+        DMEMGETPRT(&mut PrtAperture) = 0xC018800C,
+        /// Get available memory size
+        DMEMGETAVAIL(&mut DmemAvailable) = 0xC0208016,
+        /// Allocate direct memory
+        DMEMALLOC(&mut DmemAllocate) = 0xC0288001,
+        /// Query direct memory
+        DMEMQUERY(&mut DmemQuery) = 0x80288012,
```
SuchAFuriousDeath, line 122: This one looks like it should be non-mut

SuchAFuriousDeath (changes requested):

src/kernel/src/dmem/mod.rs
```diff
@@ -1,40 +1,96 @@
...
     fs: Arc<Fs>,
+    dmem: [DmemDevice; 3],
 }
 
 impl DmemManager {
+    const DMEM_TOTAL_SIZE: usize = 0x13C_000_000;
+
     pub fn new(fs: &Arc<Fs>, sys: &mut Syscalls) -> Arc<Self> {
-        let dmem = Arc::new(Self { fs: fs.clone() });
+        let dmem_devices: [DmemDevice; 3] = [0 as usize, 1, 2]
```
SuchAFuriousDeath, line 40: You could do an array of DmemContainer directly, the get the index by calling .enumerate(),

SuchAFuriousDeath (changes requested):

src/kernel/src/dmem/mod.rs
SuchAFuriousDeath: Although, I'm not sure if that can be turned back into an array. Maybe it can't really be created in a loop.

mikusp: It can't :c

SuchAFuriousDeath: Yeah, I looked it up, this doesn't use iterators :(

SuchAFuriousDeath: I would just consider making each separately and having them as fields separately rather than an array.

src/kernel/src/dmem/mod.rs
```diff
@@ -1,40 +1,138 @@
...
+                Gid::ROOT,
+                Mode::new(0o777).unwrap(),
+                None,
+                MakeDevFlags::empty(),
+            ) {
+                Ok(v) => Ok(DmemDevice::new(name, v)),
+                Err(e) => Err(DmemManagerInitError::CreateDmemFailed(name, e)),
+            }
+        }
+        .expect("error creating dmem0");
```
SuchAFuriousDeath, line 60: Create a DmemInitError here or something like that and use ?; It will also require to create the corresponding variant of KernelError in main.rs.

SuchAFuriousDeath (changes requested):

src/kernel/src/dmem/mod.rs
```diff
@@ -1,40 +1,138 @@
...
 mod blockpool;
 
+#[derive(Debug)]
+pub struct DmemDevice {
+    name: &'static str,
+    dev: Arc<CharacterDevice>,
+}
+
+impl DmemDevice {
+    pub const NAMES: [&'static str; 3] = ["dmem0", "dmem1", "dmem2"];
```
SuchAFuriousDeath, line 24: Is this still used anywhere?

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath (approved):

PR #789, opened 2024-03-31
Implements VFile read

SuchAFuriousDeath:


mikusp (approved):

src/kernel/src/fs/host/file.rs
```diff
@@ -256,7 +256,7 @@ impl HostFile {
                 null_mut(),
                 FILE_ATTRIBUTE_DIRECTORY,
                 FILE_SHARE_READ | FILE_SHARE_WRITE,
-                FILE_CREATE,
```
ultimaweapon, line 259: @SuchAFuriousDeath what happened here?

SuchAFuriousDeath: I'm not sure.

SuchAFuriousDeath: I think I put it there temporarily to fix an error, but the real problem was that the error should happend, but there should be a more top-level check for whether the error is EEXIST and I put it there and forgot to remove it. Something like that, probably

PR #793, opened 2024-04-01
Fixes offset related code

SuchAFuriousDeath:
This PR:
- fixes a very dumb mistake in UioMut::write_with
- wraps file offset in a Gutex
- adjusts some i/o-related syscalls.

I'm still unsure whether we want to have the offset be a prt of Uio and UioMut or whether all functions should take them as a standalone argument. In order to construct Uio/UioMut properly (without uninitialized memory or fields that are set and then ignored) we would have to pass around a closure that takes and offset and only then constructs the Uio/UioMut, which is doable, but it results in some things happening in reverse order (we would lock the offset field on file and only then call copyin, which could fail - so we would have unnecessarilly locked the file offset and then fail because of something that could've been checked before)

We could create an intermediate struct that doesn't contain the offset of course, but that feels like a bit too many structs, as we already have Uio and UioMut, which is just one struct in FreeBSD,

### DIFF
--- a/src/kernel/src/fs/file.rs
+++ b/src/kernel/src/fs/file.rs
@@ -8,6 +8,7 @@ use crate::net::Socket;
 use crate::process::VThread;
 use crate::shm::SharedMemory;
 use bitflags::bitflags;
+use gmtx::{Gutex, GutexGroup};
 use macros::Errno;
 use std::fmt::Debug;
 use std::io::{ErrorKind, Read, Seek, SeekFrom};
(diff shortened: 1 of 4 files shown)


src/kernel/src/fs/file.rs
```diff
@@ -164,13 +157,13 @@ impl Seek for VFile {
     }
 
     fn rewind(&mut self) -> std::io::Result<()> {
```
mikusp, line 159: Unnecessarily mut self?

src/kernel/src/fs/file.rs
```diff
@@ -164,13 +157,13 @@ impl Seek for VFile {
...
-        self.offset = 0;
+        *self.offset.write() = 0;
 
         Ok(())
     }
 
     fn stream_position(&mut self) -> std::io::Result<u64> {
```
mikusp, line 165: Also here?

mikusp (approved):

PR #805, opened 2024-04-06
Uses process credential on main thread

ultimaweapon:
This should be a correct one. I don't think the PS4 will allow the application main thread to use SceSysCore credential because it is a privileged one. We can find the correct answer without guessing in the function that do thread switching but the problem is I don't know where it is.

### DIFF
--- a/src/kernel/src/main.rs
+++ b/src/kernel/src/main.rs
@@ -1,7 +1,10 @@
 use crate::arch::MachDep;
 use crate::budget::{Budget, BudgetManager, ProcType};
-use crate::dev::{DebugManager, DipswManager, TtyManager};
-use crate::dmem::DmemManager;
+use crate::dev::{
+    DebugManager, DebugManagerInitError, DipswInitError, DipswManager, TtyManager,
+    TtyManagerInitError,
+};
+use crate::dmem::{DmemManager, DmemManagerInitError};
 use crate::ee::native::NativeEngine;
 use crate::ee::EntryArg;
 use crate::errno::EEXIST;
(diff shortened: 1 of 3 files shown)


SuchAFuriousDeath (approved):

ultimaweapon:
> Now we instantly fail on syscall 147 :(

The good news about this is it is a correct path for the games.

SuchAFuriousDeath:
image

ultimaweapon:
> image

Strange. Why you got int44. For me it is end up in sys_write.

SuchAFuriousDeath:
And what happens after you implement it? Or rather, after you fake it and just return the len right away?

SuchAFuriousDeath:
"Scepthread: Fatal error 'Can't set session id to 2 (errno = 1)'" is what it's trying to write for me btw

SuchAFuriousDeath:
> > image
> 
> Strange. Why you got int44. For me it is end up in sys_write.

Sorry btw, I myself didn't realize that I had sys_write stubbed. I merged main into my WIP branch. By the way, we should be thinking about sys_mmap_dmem, that seems to be the main obstacle that's ahead of us. Mikusp said he already has some ideas.

ultimaweapon:
I'll let you guys handle that. What I'm currently focus on is moving the application to run inside a VM.

SuchAFuriousDeath:
Where do we even begin? 👀

ultimaweapon:
Wait, you said @mikusp already have some ideas.

SuchAFuriousDeath:
He has some ideas for dmem, but not for the privilege issue, that's what I was talking about.

ultimaweapon:
> > image
> 
> Strange. Why you got int44. For me it is end up in sys_write.

You mean you got int44 after this?

SuchAFuriousDeath:
yup

SuchAFuriousDeath:
I guess I didn't actually say that 😅 Yeah, I get it right after. And from the description, it seems that it's the sys_setid. However, that means that is_system has to be true, doesn't it? 🤔

ultimaweapon:
> However, that means that is_system has to be true, doesn't it?

It should not. I'm confused why it raise int44. It seems like the PS4 expect the process to be a system process somehow.

PR #810, opened 2024-04-06
Fixes incorrect process ID for game process

### LINKED ISSUE #809: Find out why switching the main thread credential to a game credential cause int44
This caused by #805. The problem is I believed #805 already did the right thing but somehow the `libkernel.sprx` expect the main thread to have a system credential, which should not because I don't think the PS4 will run the game process with a privileged credential.

ultimaweapon:
Even though I was already known PID 1 is special in *nix world when I wrote this code but I still use it.

Fixes #809.


SuchAFuriousDeath:
Whoa

SuchAFuriousDeath (approved):

PR #811, opened 2024-04-06
Implements sys_netcontrol, refactors Filedesc::alloc

SuchAFuriousDeath:


SuchAFuriousDeath:
I use arnd here instead of the actual random generator the PS4 uses

src/kernel/src/dev/dmem.rs
```diff
@@ -71,7 +71,7 @@ impl DeviceDriver for Dmem {
             // TODO: properly implement this
             IoCmd::DMEMTOTAL(size) => *size = self.total_size,
             IoCmd::DMEMGETPRT(_prt) => todo!(),
-            IoCmd::DMEMGETAVAIL(_avail) => todo!(),
+            IoCmd::DMEMGETAVAIL(_avail) => {}
```
ultimaweapon, line 74: Why remove `todo` here?

ultimaweapon (changes requested):

src/kernel/src/dev/dmem.rs
SuchAFuriousDeath: I put it back

SuchAFuriousDeath: It seems to work without it, but we probably shouldn't play around with fire.

SuchAFuriousDeath:
@ultimaweapon Am I crazy, or is even sys_revoke and TIOCSCTTY not called now after the proc id change?

ultimaweapon:
> @ultimaweapon Am I crazy, or is even sys_revoke and the TIOCSCTTY ioctl not called now after the proc id change?

That make senses.

src/kernel/src/net/mod.rs
```diff
@@ -132,7 +173,7 @@ impl NetManager {
     }
 
     fn sys_socketex(self: &Arc<Self>, td: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
-        let name = unsafe { i.args[0].to_str(32)? };
+        let name = unsafe { i.args[0].to_str(32)? }.unwrap();
```
ultimaweapon, line 176: Is not name optional?

ultimaweapon (changes requested):

src/kernel/src/net/mod.rs
SuchAFuriousDeath: yup

src/kernel/src/net/mod.rs
```diff
@@ -77,6 +78,45 @@ impl NetManager {
...
+        info!("Netcontrol called with op = {op}.");
+
+        let buf = if buf.is_null() {
+            None
+        } else {
+            if buflen > 160 {
+                return Err(SysErr::Raw(EINVAL));
+            }
+
+            Some(unsafe { core::slice::from_raw_parts_mut(buf, buflen as usize) })
```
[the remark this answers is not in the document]
SuchAFuriousDeath: And later, there is a similar check for copyout. But since we take the original ref and don't copy anything, I don't know what to do with that.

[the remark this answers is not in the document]
SuchAFuriousDeath: It's the same thing as sys_ioctl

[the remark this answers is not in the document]
ultimaweapon: > And later, there is a similar check for copyout. But since we take the original ref and don't copy anything, I don't know what to do with that.

The problem here is our behavior is difference than the PS4. On the PS4 if `0x30000000` does not set on `op` it will not read the input buffer at all. That mean it will ignore any input but our implementation is not.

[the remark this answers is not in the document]
SuchAFuriousDeath: Refactored. Check it out.

ultimaweapon (approved):

PR #819, opened 2024-04-08
Implements sockets, initializes rng, gc and camera devices

SuchAFuriousDeath:
As I write in the doc comment for 'SocketBackend', the protocol implementation isn't exactly ideal and will probably change a lot before we settle on something. I consider it sufficient for now.

### DIFF
--- a/src/kernel/src/dev/gc.rs
+++ b/src/kernel/src/dev/gc.rs
@@ -1,12 +1,23 @@
 use crate::{
     errno::Errno,
-    fs::{CharacterDevice, DeviceDriver, IoCmd, OpenFlags},
+    fs::{
+        make_dev, CharacterDevice, DeviceDriver, DriverFlags, IoCmd, MakeDevError, MakeDevFlags,
+        Mode, OpenFlags,
+    },
     process::VThread,
+    ucred::{Gid, Uid},
 };
 use std::sync::Arc;
+use thiserror::Error;
 
 #[derive(Debug)]
 struct Gc {}
+
+impl Gc {
+    fn new() -> Self {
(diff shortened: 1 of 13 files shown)


SuchAFuriousDeath:
One has to commit a lot of programming war crimes to reach all this.
Some things I skipped on the way include mmap_dmem, mprotect, munmap, mdb_service and clock_gettime.

About the rng algorithms: Once again, red-prig just uses arnd here, so it's probaly fine for us to use it as well, at least for now
About the socket ioctl: It's in-only, the socket is just opened, the ioctl is ran and the socket is closed and never ran again. I don't ink we want to spend time on this.

src/kernel/src/dev/gc.rs
```diff
@@ -20,13 +31,48 @@ impl DeviceDriver for Gc {
...
         cmd: IoCmd,
-        td: Option<&VThread>,
+        _: Option<&VThread>,
     ) -> Result<(), Box<dyn Errno>> {
-        todo!()
+        match cmd {
+            IoCmd::GC12(_) => todo!("GC12 ioctl"),
+            IoCmd::GC16(_) => todo!("GC16 ioctl"),
+            IoCmd::GC25(_) => todo!("GC25 ioctl"),
+            IoCmd::GC27(_) => todo!("GC32 ioctl"),
```
VocalFan, line 44: Change todo message from GC32 to GC27.

VocalFan (commented):
Get reviewed (nerd)

mikusp (approved):

mikusp (approved):

PR #820, opened 2024-04-11
Add sys_mdbg_service case 01

VocalFan:
This PR adds sys_mdbg_service or, specifically, case 01 which requests for a namedobj to be created using args 1 onwards. To make this work, this also refactors the way namedobj_create is called.


VocalFan:
@ultimaweapon May I request judgement here? XD

src/kernel/src/namedobj/mod.rs
```diff
@@ -11,17 +11,20 @@ impl NamedObjManager {
     pub fn new(sys: &mut Syscalls) -> Arc<Self> {
         let namedobj = Arc::new(Self {});
 
-        sys.register(557, &namedobj, Self::sys_namedobj_create);
+        sys.register(557, &Arc::clone(&namedobj), Self::sys_namedobj_create);
```
SuchAFuriousDeath, line 14: tf is this? :eyes:

SuchAFuriousDeath (changes requested):

src/kernel/src/namedobj/mod.rs
VocalFan: Without cloning, it sort of panics in the register-

VocalFan: @SuchAFuriousDeath If I can find a way to make namedobj_create a public function, then I'll move mdbg.

SuchAFuriousDeath: > Without cloning, it sort of panics in the register-

I don't believe that. Everywhere else it's fine.

VocalFan: Nevermind, I think I did that as a pseudo fix to an earlier bug... Seems to work fine without cloning.

mikusp (approved):

src/kernel/src/namedobj/mod.rs
```diff
@@ -42,6 +45,38 @@ impl NamedObjManager {
...
+
+        self.namedobj_create(td, name, data, flags)
+    }
+
+    fn sys_mdbg_service(self: &Arc<Self>, td: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
+        // Get the mdbg arg
+        let mdbg_type: i32 = i.args[0].try_into().unwrap();
+
+        match mdbg_type {
+            0x1 => {
```
SuchAFuriousDeath, line 67: If you look into ghidra, you'll see that this is wrong.
This whole chucnk is missing.
image

SuchAFuriousDeath (changes requested):

PR #830, opened 2024-04-13
Optimize m_games

VocalFan:
I noticed an issue where any event that updated the m_games QListView would cause some lag.

Scrolling was very snappy/laggy, resizing the window would cause a delay and lag before the QListView would fit the resized window, and going from the Logs tab to the Games tab had a delay before the Games tab displayed.

I figured out that the issue was all the Icon0.pngs being reloaded EVERY time the QListView was updated. After adding a cache of the Icons, these events mentioned before are now smooth and with reduced CPU usage.
Before:
5% cpu usage when going from Logs to Games with a noticable 500ms~ delay before Games tab being displayed. (used stopwatch.)

After:
2% cpu usage when going from Logs to Games with instant feedback.

### DIFF
--- a/src/game_models.cpp
+++ b/src/game_models.cpp
@@ -13,29 +13,30 @@ Game::~Game()
 {
 }
 
-QPixmap Game::icon() const
-{
+QPixmap Game::icon() const {
+    if (!m_cachedIcon.isNull()) {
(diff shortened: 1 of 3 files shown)


mikusp (approved):
LGTM

PR #834, opened 2024-04-15
Point Readme to Wiki for building steps.

VocalFan:
After a few users being confused, I decided to move the building and development guide to the Wiki.

Besides from this, I also added links to download every requirement for Windows, as Windows is less about command-based installs and more URL-download installs.

I also fixed a few typos and tried to clean up styling a bit.

With us going from a readme file to a Wiki page, the build and development steps also have an easy to use Table of Contents to click exactly to where they need instead of scrolling through to find what they need. While also making it so the ReadME isn't cluttered and confusing for non-developers.

https://github.com/obhq/obliteration/wiki/Compilation-&-Development


ultimaweapon:
@obhq/developers what do you guys think? If we are going to move important things to Wiki like this one I'll apply the same code review rules to Wiki as well.

VocalFan:
> @obhq/developers what do you guys think? If we are going to move important things to Wiki like this one I'll apply the same code review rules to Wiki as well.

Wiki makes sense for detailed things like this. Built-in Table of Contents, better formatting, keeps things organized into separate pages, and declutters the ReadME for basic users.

SuchAFuriousDeath:
Seems reasonable to me. I expect a lot more related things to pile up over time and at some point there would be too many and too niche to be in the README.

mikusp:
Agreed, that seems like a good idea

mikusp (approved):

ultimaweapon:
Seems like Wiki does not allow code review. We will use GitHub page with https://github.com/squidfunk/mkdocs-material instead.

VocalFan:
I think I'm good, thank you. I'll be normal like every other emulator repository and use Wiki :P

VocalFan:
*After all, only members can edit the wiki... Which even then... Just ask me to do it and I'll get it done.*

ultimaweapon:
> *After all, only members can edit the wiki... Which even then... Just ask me to do it and I'll get it done.*

That is the reason why we need to move to something else that requires a review.

VocalFan:
I... But... It's just us??? You, me, acetone, SAFD?

VocalFan:
I mean, literally...

https://github.com/Ryujinx/Ryujinx/wiki

A group about 100 times bigger than us... They have no issues what so ever with the wiki.

All 3 of us agreed wiki was fine.

You're trying to make this harder than it has to be.

ultimaweapon:
Your answer does not solve the problem I stated. What you did is just tell me everyone uses Wiki without telling me how to solve the problem.

VocalFan:
> Your answer does not solve the problem I stated. What you did is just tell me everyone uses Wiki without telling me how to solve the problem.

Because the problem you are worried about, really shouldn't be something you're worrying about. WE. Agreed to the Wiki being used for easily accessible information! YOU. Are worried about an issue much larger groups have no issues about.

If you're pushing this so hard, then that just shows you don't trust the people you add to your members list.

ultimaweapon:
> If you're pushing this so hard, then that just shows you don't trust the people you add to your members list.

Correct, that's why I required every PR to be reviewed by the other member.

PR #836, opened 2024-04-15
sys_mlock and sys_mprotect

mikusp:


src/kernel/src/vm/mod.rs
```diff
@@ -211,12 +226,13 @@ impl Vm {
...
                 // Check if we need to split in the middle.
                 let decommit = if end < info.end() {
-                    adds.push(Alloc {
+                    adds.push(Mapping {
                         addr: end,
                         len: (info.end() as usize) - (end as usize),
                         prot: info.prot,
                         name: info.name.clone(),
                         storage: info.storage.clone(),
+                        locked: false,
```
ultimaweapon, line 235: Is not the locked state should be the same as the current mapping?

ultimaweapon (changes requested):

PR #837, opened 2024-04-16
Revises vectored I/O system

### LINKED ISSUE #817: Redesign Uio and IoVec
The problem with our current `IoVec` and `Uio` is:

- `IoVec` is constructed from a const pointer but provide a method `ptr` that return a mut pointer.
- Two copy of `Uio`, one for immutable and one for mutable.

ultimaweapon:
Resolves #817.

### DIFF
--- a/src/kernel/src/fs/file.rs
+++ b/src/kernel/src/fs/file.rs
@@ -1,12 +1,7 @@
-use super::{CharacterDevice, IoCmd, Stat, TruncateLength, Uio, UioMut, Vnode};
-use crate::dmem::BlockPool;
-use crate::errno::Errno;
-use crate::errno::{EINVAL, ENOTTY, ENXIO, EOPNOTSUPP};
+use super::{IoCmd, IoLen, IoVecMut, Stat, TruncateLength, Vnode};
+use crate::errno::{Errno, EINVAL, ENOTTY, ENXIO, EOPNOTSUPP};
 use crate::fs::{IoVec, PollEvents};
-use crate::kqueue::KernelQueue;
-use crate::net::Socket;
 use crate::process::VThread;
-use crate::shm::SharedMemory;
 use bitflags::bitflags;
 use gmtx::{Gutex, GutexGroup};
 use macros::Errno;
@@ -18,19 +13,28 @@ use thiserror::Error;
 /// An implementation of `file` structure.
 #[derive(Debug)]
 pub struct VFile {
-    ty: VFileType,      // f_type
-    flags: VFileFlags,  // f_flag
-    offset: Gutex<i64>, // f_offset
+    ty: VFileType,             // f_type
+    flags: VFileFlags,         // f_flag
+    vnode: Option<Arc<Vnode>>, // f_vnode
+    offset: Gutex<u64>,        // f_offset
+    backend: Box<dyn FileBackend>,
 }
 
 impl VFile {
-    pub fn new(ty: VFileType, flags: VFileFlags) -> Self {
+    pub fn new(
+        ty: VFileType,
+        flags: VFileFlags,
+        vnode: Option<Arc<Vnode>>,
+        backend: Box<dyn FileBackend>,
+    ) -> Self {
         let gg = GutexGroup::new();
 
         Self {
             ty,
             flags,
+            vnode,
             offset: gg.spawn(0),
+            backend,
         }
     }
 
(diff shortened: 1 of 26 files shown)


src/kernel/src/fs/uio.rs
```diff
@@ -1,163 +1,169 @@
...
+    }
+}
+
+/// Represents an error when [`IoLen`] fails to construct.
+#[derive(Debug)]
+pub struct IoLenError(());
+
+impl Error for IoLenError {}
+
+impl Errno for IoLenError {
```
SuchAFuriousDeath, line 159: Just as a sidenote, we could make Errno derivable for structs too.

src/kernel/src/fs/uio.rs
```diff
@@ -1,163 +1,169 @@
...
+    fn errno(&self) -> NonZeroI32 {
+        EINVAL
+    }
+}
+
+impl Display for IoLenError {
```
SuchAFuriousDeath, line 165: Can't this be derived with thiserror?  You just put the #[error(...)] above the struct

src/kernel/src/fs/host/file.rs
```diff
@@ -463,51 +433,91 @@ impl HostFile {
...
+                break Err(Error::from_raw_os_error(unsafe {
                     RtlNtStatusToDosError(err).try_into().unwrap()
-                }))
+                }));
             }
         }
     }
+
+    #[cfg(windows)]
+    fn directory_access_flags() -> u32 {
```
SuchAFuriousDeath, line 498: This can be const I think

src/kernel/src/fs/host/file.rs
```diff
@@ -463,51 +433,91 @@ impl HostFile {
...
+            | FILE_READ_EA
+            | FILE_WRITE_ATTRIBUTES
+            | FILE_WRITE_EA
+            | SYNCHRONIZE
+            | FILE_LIST_DIRECTORY
+            | FILE_TRAVERSE
+    }
+
+    #[cfg(windows)]
+    fn directory_options() -> u32 {
```
SuchAFuriousDeath, line 514: Same here

src/kernel/src/fs/file.rs
```diff
@@ -18,19 +13,28 @@ use thiserror::Error;
...
 #[derive(Debug)]
 pub struct VFile {
-    ty: VFileType,      // f_type
-    flags: VFileFlags,  // f_flag
-    offset: Gutex<i64>, // f_offset
+    ty: VFileType,             // f_type
+    flags: VFileFlags,         // f_flag
+    vnode: Option<Arc<Vnode>>, // f_vnode
+    offset: Gutex<u64>,        // f_offset
+    backend: Box<dyn FileBackend>,
```
SuchAFuriousDeath, line 20: I'm not sure I fully understand the decision to use dyn here. I understand it for cases where FreeBSD has just some ops + void * data, but not here, where FreeBSD has ops + data + type. Here, we technically allow inconsistent state between the VFileType and backend. I get that using enums is a bit tedious (there's a lot of matches), but we could still utilize your new VnodeFileBackend CDevFilebackend in the enum as it was before.

SuchAFuriousDeath (changes requested):

src/kernel/src/fs/file.rs
ultimaweapon: Actually on FreeBSD the `ops` + `data` is decoupled from `type`. An example is `devfs_ops_f` that use `DTYPE_VNODE` as a type.

SuchAFuriousDeath: Also, dyn is slower that enums (I'm pretty sure) which is why I don't want to use it for no good reason, which seems to me, is the case here.

src/kernel/src/fs/uio.rs
ultimaweapon: Better to create an issue for this.

src/kernel/src/fs/file.rs
SuchAFuriousDeath: > Actually on FreeBSD the `ops` + `data` is decoupled from `type`. An example is `devfs_ops_f` that use `DTYPE_VNODE` as a type.

I completely missed that :D I guess it makes sense then

ultimaweapon: The dynamic dispatch does not have performance impact in most cases. Unless we can prove the performance problem causes by a dynamic dispatch we don't need to worry about it.

SuchAFuriousDeath (approved):

src/kernel/src/fs/file.rs
ultimaweapon: Not sure if you know about this. The dynamic dispatch on trait object is just a call on a function pointer, the same as what FreeBSD use extensively on their `ops` things. That mean in worst case our performance will be on par with FreeBSD.

SuchAFuriousDeath: I do know that. I also did know that generally, the performace impact is negligible. My point was simply that I saw no compelling reason to refactor to use dyn, but you gave me one. I automatically assumed that DTYPE_DEV is used for devfs_ops_f, but it seems that it's actually only used for some Linux compatibility thing. Interesting.

PR #846, opened 2024-04-19
Fixes socket address implementation

SuchAFuriousDeath:
This replaces the unsound code with a todo!(). It also fixes some code in shm_open which was incorrect and adds some logging.

### DIFF
--- a/src/kernel/src/net/mod.rs
+++ b/src/kernel/src/net/mod.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use bitflags::bitflags;
 use macros::Errno;
+use std::fmt::Debug;
 use std::num::NonZeroI32;
 use std::sync::Arc;
 use thiserror::Error;
@@ -51,7 +52,11 @@ impl NetManager {
             MessageFlags::from_bits_retain(flags)
         };
 
-        todo!()
+        let msg = unsafe { &mut *msg };
+
+        info!("Recieving a message {msg:?} from fd {fd} with flags {flags:?}.");
+
+        Ok((msg.iovec_len as i32).into())
     }
 
     fn sys_sendmsg(self: &Arc<Self>, td: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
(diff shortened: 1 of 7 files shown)


src/kernel/src/net/mod.rs
```diff
@@ -51,7 +52,11 @@ impl NetManager {
             MessageFlags::from_bits_retain(flags)
         };
 
-        todo!()
+        let msg = unsafe { &mut *msg };
+
+        info!("Recieving a message {msg:?} from fd {fd} with flags {flags:?}.");
+
+        Ok((msg.iovec_len as i32).into())
```
ultimaweapon, line 59: This basically a stub because it response with unexpected data. Some games might handle this correctly but some games may not, which may cause segmentation fault. In the later case it can cause headache for us to find what it the cause of segmentation fault.

ultimaweapon (changes requested):

src/kernel/src/net/mod.rs
SuchAFuriousDeath: It is a stub, sorry. I accidentally left it in here.

src/kernel/src/shm/mod.rs
```diff
@@ -28,18 +28,11 @@ impl SharedMemoryManager {
         let flags: OpenFlags = i.args[1].try_into().unwrap();
         let mode: u32 = i.args[2].try_into().unwrap();
 
-        if (flags & OpenFlags::O_ACCMODE != OpenFlags::O_RDONLY)
-            || (flags & OpenFlags::O_ACCMODE != OpenFlags::O_RDWR)
-        {
+        if flags & OpenFlags::O_ACCMODE | OpenFlags::O_RDWR != OpenFlags::O_RDWR {
```
ultimaweapon, line 31: This is quite confusing. Better to change to `(flags | OpenFlags::O_RDWR) & OpenFlags::O_ACCMODE != OpenFlags::O_RDWR`.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #851, opened 2024-04-20
Implements sceSblRcMgrIsAllowULDebugger and sceSblRcMgrIsSoftwagnerQafForAcmgr

ultimaweapon:


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
@ultimaweapon By the way, do we plan to use low level apis even for audio? Just curious, it's not important.

ultimaweapon:
No idea yet. We probably need to see how the PS4 communicate with its kernel first before we can decide.

PR #852, opened 2024-04-20
Adjusts hypervisor errors on mac to use NonZero

SuchAFuriousDeath:
Let's see if this compiles.

I don't know if this is niche knowledge or not, but Rust recently switched from distinct NonZero types to using a generic trait. so we can utliize this here.

EDIT: oops, it has not been stabilized yet


SuchAFuriousDeath:
Seems like rust was updated, so this can be merged now

ultimaweapon (approved):

PR #856, opened 2024-04-22
Gets vcpu mmap size, refactors kvm and vm 

SuchAFuriousDeath:
I think this is rougly what we want. Tbh, I'm not completely sure about the Error::from_raw_os_error.

### DIFF
--- a/src/kernel/src/hv/linux/kvm.cpp
+++ b/src/kernel/src/hv/linux/kvm.cpp
@@ -67,6 +67,11 @@ extern "C" int kvm_set_user_memory_region(
     return 0;
 }
 
+extern "C" int kvm_get_vcpu_mmap_size(int kvm)
+{
+    return ioctl(kvm, KVM_GET_VCPU_MMAP_SIZE);
+}
+
 extern "C" int kvm_create_vcpu(int vm, int id, int *fd)
 {
     auto vcpu = ioctl(vm, KVM_CREATE_VCPU, id);
--- a/src/kernel/src/hv/linux/kvm.cpp
+++ b/src/kernel/src/hv/linux/kvm.cpp
@@ -89,3 +89,13 @@ extern "C" int kvm_run(int vcpu)
 {
     return ioctl(vcpu, KVM_RUN);
 }
+
+extern "C" int kvm_get_regs(int vcpu, kvm_regs *regs)
+{
+    return ioctl(vcpu, KVM_GET_REGS, regs);
+}
+
+extern "C" int kvm_set_regs(int vcpu, kvm_regs *regs)
+{
+    return ioctl(vcpu, KVM_SET_REGS, regs);
+}
--- a/src/kernel/src/hv/linux/kvm.cpp
+++ b/src/kernel/src/hv/linux/kvm.cpp
@@ -95,7 +95,7 @@ extern "C" int kvm_get_regs(int vcpu, kvm_regs *regs)
     return ioctl(vcpu, KVM_GET_REGS, regs);
 }
 
-extern "C" int kvm_set_regs(int vcpu, kvm_regs *regs)
+extern "C" int kvm_set_regs(int vcpu, kvm_regs *const regs)
 {
     return ioctl(vcpu, KVM_SET_REGS, regs);
 }
--- a/src/kernel/src/hv/linux/kvm.cpp
+++ b/src/kernel/src/hv/linux/kvm.cpp
@@ -84,3 +84,8 @@ extern "C" int kvm_create_vcpu(int vm, int id, int *fd)
 
     return 0;
 }
+
+extern "C" int kvm_run(int vcpu)
+{
+    return ioctl(vcpu, KVM_RUN);
+}
--- a/src/kernel/src/hv/linux/mod.rs
+++ b/src/kernel/src/hv/linux/mod.rs
@@ -49,7 +49,9 @@ impl Kvm {
     pub fn get_vcpu_mmap_size(&self) -> Result<usize, HypervisorError> {
         match unsafe { kvm_get_vcpu_mmap_size(self.0.as_raw_fd()) } {
             size @ 0.. => Ok(size as usize),
-            v => Err(HypervisorError::GetMmapSizeFailed(Error::from_raw_os_error(v))),
+            v => Err(HypervisorError::GetMmapSizeFailed(
+                Error::from_raw_os_error(v),
+            )),
         }
     }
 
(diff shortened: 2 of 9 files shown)


SuchAFuriousDeath:
I decided to do a refactor, so Kvm and Vm now have their own wrapper types around OwnedFd. I think this is better because it makes it more clear whata method on kvm and what is a method on vm

SuchAFuriousDeath:
@ultimaweapon If we only run the game in a vm, how do we deal with multiple threads?

ultimaweapon:
It will be the same as other kernels. That is, switch between threads when the VM exits due to the time slice used up.

SuchAFuriousDeath:
Didn't know that was possible. Nice

src/kernel/src/main.rs
```diff
@@ -594,10 +594,6 @@ struct Args {
     #[arg(long)]
     #[serde(default)]
     pro: bool,
-
-    #[arg(long)]
-    #[serde(default)]
-    idps: ConsoleId,
```
ultimaweapon, line 600: What wrong with this?

ultimaweapon (changes requested):

src/kernel/src/main.rs
SuchAFuriousDeath: Nothing, I just forgot to exclude it from the commit. I just removed it so that I can actually run the emualator, which is not possible right now, because the argument is required

SuchAFuriousDeath: Actually... now that I run the emulator again, it's fine. But I swear that previously it would complain about the argument not being specified :eyes: .

SuchAFuriousDeath:
So I guess ... now we just map the module into memory, set the registers and run the KVM_RUN ioctl, right? Roughly

SuchAFuriousDeath:
+ the scheduler

ultimaweapon:
We need to fix the process VM to allocate from `Ram` and setup page table for the game process.

src/kernel/src/hv/ram.rs
```diff
@@ -40,6 +39,11 @@ impl Ram {
             mem.cast()
         };
 
+        Ok(Self { addr, mem })
+    }
+
+    #[cfg(windows)]
+    pub fn new(addr: usize) -> Result<Self, Error> {
```
ultimaweapon, line 46: The downside with splitting the public method like this is we will have a duplicated documentation so better to leave it merged like the previous version. If you want to split it better to split it into something like `new_unix` and `new_win32` as private functions and call it from `new` instead.

ultimaweapon (changes requested):

VocalFan:
@ultimaweapon 

image

IDPS do be broken tho-

VocalFan:
Yeah you... Never defined it in UI, Ultima xD

ultimaweapon:
Better to create an issue for this.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -1,81 +1,183 @@
...
+        Ok(Self(fd))
+    }
+
+    pub fn get_vcpu_mmap_size(&self) -> Result<usize, HypervisorError> {
+        match unsafe { kvm_get_vcpu_mmap_size(self.0.as_raw_fd()) } {
+            size @ 0.. => Ok(size as usize),
+            v => Err(HypervisorError::GetMmapSizeFailed(
                 Error::from_raw_os_error(v),
-            ))
+            )),
```
ultimaweapon, line 50: I don't think this is correct.

ultimaweapon (changes requested):

src/kernel/src/hv/linux/mod.rs
SuchAFuriousDeath: Error::last_os_error() instead?

src/kernel/src/hv/linux/kvm.cpp
```diff
@@ -79,3 +84,18 @@ extern "C" int kvm_create_vcpu(int vm, int id, int *fd)
...
+{
+    return ioctl(vcpu, KVM_RUN);
+}
+
+extern "C" int kvm_get_regs(int vcpu, kvm_regs *regs)
+{
+    return ioctl(vcpu, KVM_GET_REGS, regs);
+}
+
+extern "C" int kvm_set_regs(int vcpu, kvm_regs *regs)
```
ultimaweapon, line 98: `regs` should be a const pointer here since it is const on Rust side.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -1,81 +1,181 @@
...
-    match unsafe { kvm_create_vcpu(vm.as_raw_fd(), id, &mut vcpu) } {
-        0 => Ok(unsafe { OwnedFd::from_raw_fd(vcpu) }),
-        v => Err(Error::from_raw_os_error(v)),
+impl VCpu {
+    pub fn get_regs(&self) -> Result<KvmRegs, Error> {
+        let mut regs = MaybeUninit::uninit();
+
+        match unsafe { kvm_get_regs(self.fd.as_raw_fd(), regs.as_mut_ptr()) } {
+            0 => Ok(unsafe { regs.assume_init() }),
+            v => Err(Error::from_raw_os_error(v)),
```
ultimaweapon, line 161: I don't think this correct.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -1,81 +1,181 @@
...
+        }
+    }
+
+    pub fn set_regs(&self, regs: KvmRegs) -> Result<(), Error> {
+        match unsafe { kvm_set_regs(self.fd.as_raw_fd(), &regs) } {
+            0 => Ok(()),
+            v => Err(Error::from_raw_os_error(v)),
```
ultimaweapon, line 168: Same here.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -1,81 +1,181 @@
...
 
-pub fn create_vcpu(vm: BorrowedFd, id: i32) -> Result<OwnedFd, Error> {
-    let mut vcpu = -1;
+#[derive(Debug)]
+pub struct VCpus([VCpu; 8]);
+
+#[derive(Debug)]
+struct VCpu {
+    fd: OwnedFd,
+    kvm_run: NonNull<KvmRun>,
```
ultimaweapon, line 152: Seems like a memory leak here.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -84,4 +184,23 @@ extern "C" {
...
     fn kvm_create_vcpu(vm: c_int, id: c_int, fd: *mut c_int) -> c_int;
+
+    fn kvm_run(vcpu: c_int) -> c_int;
+    fn kvm_get_regs(vcpu: c_int, regs: *mut KvmRegs) -> c_int;
+    fn kvm_set_regs(vcpu: c_int, regs: *const KvmRegs) -> c_int;
+}
+
+#[derive(Debug, Error)]
+pub enum CreateVCpusError {
+    #[error("Failed to create vCPU #{1}")]
```
ultimaweapon, line 195: Wrong convention for the message.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -84,4 +184,23 @@ extern "C" {
...
+    CreateVcpuFailed(#[source] CreateVCpuError, u8),
+}
+
+#[derive(Debug, Error)]
+pub enum CreateVCpuError {
+    #[error("Failed to create vCPU")]
```
ultimaweapon, line 201: Same here.

src/kernel/src/hv/linux/mod.rs
```diff
@@ -84,4 +184,23 @@ extern "C" {
...
+    CreateVcpuFailed(#[source] Error),
+
+    #[error("Failed to mmap KVM_RUN")]
```
ultimaweapon, line 204: Same here.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #860, opened 2024-05-02
Adds placeholder value for Version

VocalFan:
Some PS4 system apps don't have a Version value defined in Param.SFO.

This makes it where if there is no value, a placeholder is used instead.

Before:
image

After:
image

Allows PS4's UI app to display without errors.

*And before IDPS was added, was able to run in UI :P*

### DIFF
--- a/src/kernel/src/dev/camera.rs
+++ b/src/kernel/src/dev/camera.rs
@@ -77,7 +77,7 @@ impl CameraManager {
             Gid::ROOT,
             Mode::new(0o666).unwrap(),
             None,
-            MakeDevFlags::MAKEDEV_ETERNAL,
+            MakeDevFlags::ETERNAL,
         )?;
 
         Ok(Arc::new(Self { camera }))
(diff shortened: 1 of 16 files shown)


src/param/src/lib.rs
```diff
@@ -166,7 +166,7 @@ impl Param {
             content_id: content_id.ok_or(ReadError::MissingContentId)?,
             title,
             title_id: title_id.ok_or(ReadError::MissingTitleId)?,
-            version: version.ok_or(ReadError::MissingVersion)?,
```
ultimaweapon, line 169: This need to be handled on the UI side. What you need to do here is make version as an optional.

ultimaweapon (changes requested):

src/param/src/lib.rs
VocalFan: Not sure how that's UI sided when it... Gives the error

```rs
    #[error("VERSION parameter not found")]
    MissingVersion,
 ```
 
 ?
 
 Which also means even if ignored in UI, it'd crash on the kernel... When obviously the PS4 can run its own UI without the version param.

PR #863, opened 2024-05-06
Refactors Vcpu drop

SuchAFuriousDeath:


src/kernel/src/hv/linux/mod.rs
```diff
@@ -139,33 +139,32 @@ impl Vm {
...
-        use libc::munmap;
-
-        unsafe {
-            if munmap(self.kvm_run.as_ptr().cast(), self.mmap_size) < 0 {
-                panic!("failed to munmap KVM_RUN: {}", Error::last_os_error());
-            };
+        for vcpu in &self.vcpus {
+            unsafe {
+                libc::munmap(vcpu.kvm_run.as_ptr().cast(), self.mmap_size);
+            }
```
ultimaweapon, line 157: I don't think this is a good idea. Each memory map belongs to each vCPU. We will have a bug if someone move out a vCPU from this struct and doesn't realize they also need to fix this.

ultimaweapon (changes requested):

src/kernel/src/hv/linux/mod.rs
SuchAFuriousDeath: Maybe you're right. Hmmm

SuchAFuriousDeath:
I don't really have enough for a PR now

PR #865, opened 2024-05-19
Uses clap builder to parse kernel arguments

ultimaweapon:
The reason is because clap derive does not supports default value from `Default` trait (AFAIK), which we required for IDPS.

### DIFF
--- a/src/kernel/src/args.rs
+++ b/src/kernel/src/args.rs
@@ -1,0 +1,93 @@
+use crate::idps::ConsoleId;
+use clap::{command, value_parser, Arg, ArgAction};
+use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


SuchAFuriousDeath:
maybe default_value_t = Default::default?

SuchAFuriousDeath:
Also, have you tried this?

image

SuchAFuriousDeath (approved):

ultimaweapon:
> maybe default_value_t = Default::default?

Not working.

> Also, have you tried this?
> 
> image

Not tried without `=` yet but I think the builder pattern is better here because it does not require `std::fmt::Display` implementation on the type.

PR #867, opened 2024-05-22
Moves sys_get_authinfo to ProcManager

ultimaweapon:


src/kernel/src/process/mod.rs
```diff
@@ -339,6 +341,54 @@ impl ProcManager {
         Ok(v.into())
     }
 
+    fn sys_get_authinfo(self: &Arc<Self>, td: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
+        // Get arguments.
+        let pid: i32 = i.args[0].try_into().unwrap();
```
SuchAFuriousDeath, line 346: Couldn't you convert it directly to a Option\<NonZeroI32\>?

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath (approved):

PR #869, opened 2024-05-26
Adds lifetime to IoCmd, init ipmi commands

SuchAFuriousDeath:


SuchAFuriousDeath:
@ultimaweapon Does it make sense for a each process to have it's own instance of Vm? shouldn't it be inside of the ProcManager?

ultimaweapon:
That is not possible on macOS. It also seems like not a good idea since the amount of process may be hundred.

SuchAFuriousDeath:
If it doesn't make sense... why is it that way in the code? :eyes:

SuchAFuriousDeath:
Well... I suppose it made sense with only one process

src/kernel/src/fs/ioctl.rs
```diff
@@ -91,7 +91,7 @@ macro_rules! commands {
 // TODO: implement void ioctl commands with int data
 
 commands! {
-    pub enum IoCmd {
+    pub enum IoCmd<'a> {
```
ultimaweapon, line 94: This lifetime need to be referenced on each variant too.

ultimaweapon (changes requested):

ultimaweapon:
> If it doesn't make sense... why is it that way in the code? 👀

What do you mean by that? Our code current have a single VM that will be hosting multiple processes.

SuchAFuriousDeath:
I mean the Vm struct

ultimaweapon:
In that case yes, each process will have only one. I believed the reason FreeBSD put this as a top-level is because the VM can live alone without a process. The process is just one of a thing that depend on the VM. So I think it is better to keep it on a top-level.

src/kernel/src/process/mod.rs
```diff
@@ -382,16 +382,19 @@ impl ProcManager {
...
     fn sys_cpuset_setaffinity(self: &Arc<Self>, _: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
+        let level: CpuLevel = i.args[0].try_into()?;
+        let which: CpuWhich = i.args[1].try_into()?;
+        let id: i64 = i.args[2].into();
         let cpusetsize: usize = i.args[3].into();
+        let mask: *const u8 = i.args[4].into();
 
-        // TODO: Refactor this for readability.
-        if cpusetsize.wrapping_sub(8) > 8 {
+        if cpusetsize.wrapping_sub(8) > std::mem::size_of::<CpuSet>() {
```
ultimaweapon, line 391: Our `CpuSet` is not `cpuset_t` so this is not correct.

src/kernel/src/process/mod.rs
```diff
@@ -346,7 +346,7 @@ impl ProcManager {
         let mask: *mut u8 = i.args[4].into();
 
         // TODO: Refactor this for readability.
-        if cpusetsize.wrapping_sub(8) > 8 {
+        if cpusetsize.wrapping_sub(8) > std::mem::size_of::<CpuSet>() {
```
ultimaweapon, line 349: Same here.

src/kernel/src/ipmi/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+impl IpmiManager {
+    pub fn new(syscalls: &mut Syscalls) -> Arc<Self> {
+        let ipmi = Arc::new(Self {});
+
+        syscalls.register(622, &ipmi, Self::ipmi_mgr_call);
+
+        ipmi
+    }
+
+    fn ipmi_mgr_call(self: &Arc<Self>, td: &VThread, i: &SysIn) -> Result<SysOut, SysErr> {
```
ultimaweapon, line 24: All syscall handlers should have the same naming convention.

SuchAFuriousDeath:
I'm not sure I can make us run in a VM. Maybe you have to do it

src/kernel/src/vm/mod.rs
```diff
@@ -622,7 +622,7 @@ impl Vm {
             return Err(SysErr::Raw(EINVAL));
         }
 
-        let slice_size = num_of_ops.try_into().ok().ok_or(SysErr::Raw(EINVAL))?;
+        let slice_size = num_of_ops.try_into().map_err(|_| SysErr::Raw(EINVAL))?;
```
ultimaweapon, line 625: I can't find this check on the PS4.

ultimaweapon (changes requested):

src/kernel/src/vm/mod.rs
SuchAFuriousDeath: image

I refactored it

ultimaweapon (approved):

PR #875, opened 2024-06-16
Initializes VmMgr

ultimaweapon:
The recommended way to review this PR is to review each commit separately.


SuchAFuriousDeath:
I assume the next step is to move all the memory syscalls into the new VmMgr, right?

ultimaweapon:
Correct.

SuchAFuriousDeath:
@ultimaweapon In which struct should the map of the addresses to the allocations be?

PR #879, opened 2024-06-22
Moves vm syscalls to VmMgr

SuchAFuriousDeath:


src/kernel/src/process/proc.rs
```diff
@@ -64,14 +64,15 @@ impl VProc {
 
         let gg = GutexGroup::new();
         let limits = Limits::load()?;
-        let vm = Vm::new(&mut sys)?;
```
ultimaweapon, line 67: `sys` is no longer needed here.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #891, opened 2024-07-16
Initializes AArch64 supports

ultimaweapon:
This is not going to be an official support in a near future due to we need someone to work on a dynamic recompiler. The reason of this PR is I want to be able to do some works on my MacBook AArch64 because it is inconvenience for me to access my x86 PC on a weekday after my full-time job.


ultimaweapon:
I'm surprising it can run without any fixes!
Screenshot 2024-07-17 at 01 36 34

SuchAFuriousDeath (approved):

PR #904, opened 2024-07-28
Runs the kernel

ultimaweapon:
The execution finally reach the kernel entry point!


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Amazing. By the way, I think you should know this:
image

ultimaweapon:
The problem is I don't use Intel. I just created an issue for this #905.

Issue #905, opened 2024-07-28
Make VMM support Intel CPU


ultimaweapon:
See https://github.com/obhq/obliteration/pull/904#issuecomment-2254501621 for the details. The problem is I don't use Intel CPU so I can't really test this. Need someone who have Intel CPU to work on this (or someone willing to be a tester).

ultimaweapon:
@SuchAFuriousDeath I totally forgot that I have one PC with an Intel CPU so I'll take on this issue.

ultimaweapon:
https://github.com/cloud-hypervisor/cloud-hypervisor/issues/305

ultimaweapon:
Seems like TSS is not a culprit here because TSS address used only on real-mode: https://github.com/torvalds/linux/blob/master/arch/x86/kvm/vmx/vmx.c#L3128

SuchAFuriousDeath:
Does running the vm give you the same error?

ultimaweapon:
I got the same problem as you on Intel® Core™ i5-1035G4. Trying to figure out what is the problem.

SuchAFuriousDeath:
Maybe you tried already, but the KVM_GET_VCPU_EVENTS ioctl should give some more information about what is happening.

ultimaweapon:
It turn out Intel violate the x86-64 specs on CS.

PR #909, opened 2024-07-28
Implements run and states for x86 MacOS

SuchAFuriousDeath:


SuchAFuriousDeath:
Not sure what's up with the Windows build

SuchAFuriousDeath:
It seems that we can't access the EFER register, but I assume the features we are interested in are already turned on?.

SuchAFuriousDeath:
Not sure what to do about set_cs :(

SuchAFuriousDeath:
I assume we're interested in ARM Linux and ARM Windows as well?

ultimaweapon:
Seems like the Hypervisor Framework on macOS is the worst. A lot of restrictions.

> I assume we're interested in ARM Linux and ARM Windows as well?

If someone wiling to work on this.

ultimaweapon:
BTW I'm surprised that WHP has a very good API.

VocalFan:
> BTW I'm surprised that WHP has a very good API.

Microsoft takes pride in Hyper-V, so... Makes sense.

SuchAFuriousDeath:
I think the segment registers can be set and read with this:
https://developer.apple.com/documentation/hypervisor/1441144-hv_vmx_vcpu_write_vmcs?language=objc

Here is a list of field ids. The segment registers do appear more than once. For example, for cs there's, VMCS_GUEST_CS, VMCS_HOST_CS, VMCS_GUEST_CS_LIMIT and VMCS_GUEST_CS_AR

https://developer.apple.com/documentation/hypervisor/1469436-virtual_machine_control_structur?language=objc

ultimaweapon:
Should be `VMCS_GUEST_CS`. Not sure what the purpose of the other ones.

SuchAFuriousDeath:
I guess the contents are the same as .Segment.Anonymous.Attributes on Windows? :eyes:

ultimaweapon:
> I guess the contents are the same as .Segment.Anonymous.Attributes on Windows? 👀

This really implementation specific. Maybe finding some example on the internet should be easier if the docs does not help?

ultimaweapon:
Just found a very good example: https://github.com/utshina/AHF-simple

SuchAFuriousDeath:
A good example indeed. I managed to implemente exit_reason for x86 mac because of it

SuchAFuriousDeath:
By the way the vm stuff seems to work correctly for now on my Linux machine.

ultimaweapon:
No.

SuchAFuriousDeath:
Oh, Where does it stop for you?

ultimaweapon:
It is never stop because of `hlt`. In other words, it work as expected. Did you try on the latest `main` of from your branch?

SuchAFuriousDeath:
Both, In main, I panic here 
https://github.com/obhq/obliteration/blob/573eca32403983bd2b4c6cafda9ed9630a3362b5/src/core/src/vmm/mod.rs#L615-L617
In my branch, i the same things happens in a different place, but it is equivalent.

ultimaweapon:
The problem may be caused by a different Linux version or different CPU. What is your Linux version and CPU model? I was tested this with Arch on Ryzen 5600G.

SuchAFuriousDeath:
EndeavourOS (Which is Arch-based), kernel version 6.9.7, cpu 11th Gen Intel(R) Core(TM) i5-11400H @ 2.70GHz

ultimaweapon:
I suspect this #905 or https://github.com/firecracker-microvm/firecracker/issues/668 if your Linux running inside a VM.

SuchAFuriousDeath:
I completely forgot that the issue applies to me :joy:  That might very well be it.

SuchAFuriousDeath:
No nested VMs for me though.

SuchAFuriousDeath:
> 

It doesn't seem to be it

ultimaweapon:
Strange. At least on Windows with Ryzen it is working according to https://github.com/obhq/obliteration/pull/916#issue-2443300798

ultimaweapon:
Can we ask another people who are using Linux with AMD to test this? If it is working I'm pretty sure it is because of #905.

SuchAFuriousDeath:
What address should I set fr the TSS?

ultimaweapon:
I don't know but any virtual address need to be mapped to physical address via page table otherwise the CPU will trigger page fault, which we currently does not setup a handler to handle it.

SuchAFuriousDeath:
But the TSS that we have to set is a physical address, no?

ultimaweapon:
I don't know. You should take a look on Intel manual. But AFAIK the only place that need to be a physical address is page tables and `cr3`.

SuchAFuriousDeath:
image
It seems that I can get some information about the exit

ultimaweapon:
BTW you should not make this PR contain too many things.

SuchAFuriousDeath:
The exception is 0x80000021, the error code is 0.

ultimaweapon:
We are not alone: https://forum.proxmox.com/threads/vm-shutdown-kvm-entry-failed-hardware-error-0x80000021.109410/

ultimaweapon:
In our case I suspect TSS because `0x80000021` mean invalid guest state.

src/core/src/vmm/cpu.rs
```diff
@@ -56,19 +56,23 @@ pub trait CpuStates {
...
 }
 
-/// Contains information when a VM exited because of I/O instructions.
-#[cfg(target_arch = "x86_64")]
-pub enum CpuIo<'a> {
-    Out(u16, &'a [u8]),
+#[derive(Debug)]
+pub enum ExitReason<'a> {
+    Hlt,
+    Other,
```
ultimaweapon, line 76: We we need this?

src/core/src/vmm/mod.rs
```diff
@@ -607,16 +607,14 @@ fn run_cpu(mut cpu: impl Cpu, args: &CpuArgs) {
...
+        match exit.reason() {
+            ExitReason::Hlt => {}
+            ExitReason::IoOut(0, data) => {
+                logs.extend_from_slice(data);
+                parse_logs(&args.logs, &mut logs);
             }
-        } else if !exit.is_hlt() {
-            todo!()
+            ExitReason::IoOut(_, _) => todo!(),
+            reason => todo!("unhandled VM exit reason: {reason:?}"),
```
ultimaweapon, line 617: This should not needed here. It should already handled by CPU implementation.

ultimaweapon (changes requested):

src/core/src/vmm/cpu.rs
SuchAFuriousDeath: I'll revert it

ultimaweapon (approved):

PR #910, opened 2024-07-29
Initializes kernel logging

ultimaweapon:


SuchAFuriousDeath:
Wouldn't it be better to create an ExitReason enum instead of having several is_* functions? Also, is_* functions return booleans.

ultimaweapon:
The reason with this implementation is to not cause too many conflicts to your PR. If you are okay with some conflicts I'll replace the whole CpuExit with enum.

SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Good point. I'll rebase my PR

PR #913, opened 2024-07-31
Initializes screen implementation

ultimaweapon:


ArbestRi02:
Screen implementation seems promising.

ultimaweapon:
It is still a long road for us to release the first version. We need to migrate the user-mode kernel to the kernel mode first.

PR #916, opened 2024-08-01
This somehow fixes #914

### LINKED ISSUE #914: WinHvPlatform.dll crashes Obliteration when pressing Start on any game.
Faulting application name: Obliteration.exe, version: 0.0.1.0, time stamp: 0x66aa766c
Faulting module name: WinHvPlatform.dll, version: 10.0.22621.3672, time stamp: 0xb628bb33
Exception code: 0xc0000005
Fault offset: 0x00000000000076d7
Faulting application path: C:\Users\ghost\Downloads\obliteration-win-x64\Obliteration.exe
Faulting module path: C:\WINDOWS\SYSTEM32\WinHvPlatform.dll

VocalFan:
I don't know why a println! fixes it, I don't want to know why a println! fixes it, but it does.

Now hitting this:

```rs
    #[cfg(target_arch = "x86_64")]
    fn is_io(&mut self) -> Option<CpuIo> {
        todo!()
    }
```

### DIFF
--- a/src/core.h
+++ b/src/core.h
@@ -72,11 +72,9 @@ struct RustError *update_firmware(const char *root,
                                   void *cx,
                                   void (*status)(const char*, uint64_t, uint64_t, void*));
 
-struct Vmm *vmm_new(size_t screen, struct RustError **err);
-
 void vmm_free(struct Vmm *vmm);
 
-struct RustError *vmm_run(struct Vmm *vmm, const char *kernel);
+struct Vmm *vmm_run(const char *kernel, size_t screen, struct RustError **err);
 
 struct RustError *vmm_draw(struct Vmm *vmm);
 
--- a/src/core/src/vmm/hw/ram/builder.rs
+++ b/src/core/src/vmm/hw/ram/builder.rs
@@ -1,0 +1,465 @@
+use super::{Ram, RamError};
+use crate::vmm::VmmError;
+use obconf::BootEnv;
+use std::ops::Range;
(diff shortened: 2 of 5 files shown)


VocalFan:
Gonna fix this properly, I think just .clone might make the value live long enough.

VocalFan:
Okay so uh, if I do a clone above
`        let mut values: [WHV_REGISTER_VALUE; REGISTERS] = unsafe { zeroed() };` it works, but if I do it BELOW? It breaks...

VocalFan:
OKay I hate this, but cloning imminent.

VocalFan:
@ultimaweapon Enjoy my goofy fix... I tried putting my clone inside WHvGetVirtualProcessorRegisters, but it still errored, so... Yeah, cloning before using works... Seems like a use-after-free issue that I kind of worked around with this.

ultimaweapon:
This is not the right way to fix this. We need to figure out the reason why it crashes first. Once we know the reason we will know how to fix this properly.

VocalFan:
> This is not the right way to fix this. We need to figure out the reason why it crashes first. Once we know the reason we will know how to fix this properly.

Again, this is mainly because it's a use-after-free issue, possibly due to our abuse of unsafe code confusing the compiler... Best suggestion would be to try to use Miri and see if it detects where the cause of this use-after-free occurs.

VocalFan:
Basically, self.part gets dropped before the API call can use it, causing a memory access violation.

Issue #937, opened 2024-08-19
0xc0000409 exception on Windows


VocalFan:
Unknown what is exactly causing it as I'm stuck with using the release builds rn haha.

```
Faulting application name: Obliteration.exe, version: 0.0.1.0, time stamp: 0x66c39537
Faulting module name: Obliteration.exe, version: 0.0.1.0, time stamp: 0x66c39537
Exception code: 0xc0000409
Fault offset: 0x0000000000053649
```

Error lookup:
```
# for hex 0xc0000409 / decimal -1073740791
  STATUS_STACK_BUFFER_OVERRUN                                    ntstatus.h
# The system detected an overrun of a stack-based buffer in
# this application. This overrun could potentially allow a
# malicious user to gain control of this application.
```

ultimaweapon:
What Actions you downloaded the binary? It must be the same one that is causing this error otherwise the offset will not match.

VocalFan:
Latest Release action, Windows x64

ultimaweapon:
This is Rust panic. Can you try running Obliteration.exe from Windows Terminal? It should print something before causing this error.

VocalFan:
> This is Rust panic. Can you try running Obliteration.exe from Windows Terminal? It should print something before causing this error.

Yeah... No.

```
PS C:\Users\ghost\Downloads\obliteration-win-x64 (5)> .\Obliteration.exe
PS C:\Users\ghost\Downloads\obliteration-win-x64 (5)>
```

VocalFan:
Unless this is happening because the wizard install window appears, which if so... How the hell do I make that go away? XD I always just feed it a blank file.

ultimaweapon:
Try running https://learn.microsoft.com/en-us/sysinternals/downloads/debugview then start Obliteration.

VocalFan:
Yeah, nope... Let me  see what I can do.

```
[24760] [OBS] 
[24760] graphics-hook.dll loaded against process: Obliteration.exe
[24760] 
[24760] [OBS] 
[24760] (half life scientist) everything..  seems to be in order
```
Have this funny from OBS.

VocalFan:
```
(26c8.56b0): Security check failure or stack buffer overrun - code c0000409 (!!! second chance !!!)
Subcode: 0x7 FAST_FAIL_FATAL_APP_EXIT 
Obliteration+0x53649:
00007ff7`bd1c3649 cd29            int     29h
```

VocalFan:
```
[...]

PR #938, opened 2024-08-20
Hooks Rust panic with QMessageBox

ultimaweapon:
So we won't have the same problem as #937.


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Works like a charm

PR #973, opened 2024-09-08
Initializes malloc stats

ultimaweapon:


src/obkrnl/src/malloc/stage2.rs
```diff
@@ -54,7 +56,14 @@ impl Stage2 {
             zones
         });
 
-        Self { zones }
+        // TODO: Is there a better way than this?
```
SuchAFuriousDeath, line 59: std::vec::from_elem

SuchAFuriousDeath (changes requested):

src/obkrnl/src/malloc/stage2.rs
SuchAFuriousDeath: But that's only for Clone types

ultimaweapon: Not sure where did you get this from since I can't find this method on both `std` and `alloc`.

ultimaweapon: > But that's only for types that implement Clone

This is a problem since `Stats` cannot be cloned due to atomic fields.

SuchAFuriousDeath (approved):

PR #1000, opened 2024-09-28
Fixes KVM_CAP_ONE_REG is not supported on x86-64

ultimaweapon:
Although the docs said it is available on all architectures but it does not available on Arch Linux somehow.


SuchAFuriousDeath (approved):

VocalFan:
It's beautiful, we've reached the 1000s!

PR #1016, opened 2024-10-05
Initializes Target implementation for gdbstub

ultimaweapon:


gui/src/vmm/debug/aarch64.rs
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+use gdbstub::target::ext::base::BaseOps;
+use thiserror::Error;
+
+/// Implementation of [`gdbstub::target::Target`] for x86-64.
```
SuchAFuriousDeath, line 5: The platform here is wrong just as in all the other places in this file

SuchAFuriousDeath (changes requested):

SuchAFuriousDeath (approved):

PR #1021, opened 2024-10-06
Update from macos-12 to macos-13

VocalFan:
This is due to macos-12 soon being deprecated: https://github.com/actions/runner-images/issues/10721

As such, I bumped our runners that are using macos-12 to macos-13. Not macos-14, as that is an ARM64 based runner as shown here:
```
macOS 13	macos-13 or macos-13-large	macOS-13
macOS 13 Arm64	macos-13-xlarge			macOS-13-arm64
```
As long as macos-13 is used and not macos-13-xlarge, we'll stay on the intel platform.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
     uses: ./.github/workflows/ci-mac.yml
     with:
       name: Mac Intel
-      macos: macos-12
+      macos: macos-13
       kernel-target: x86_64-unknown-none
       artifact-name: obliteration-mac-intel
   build-mac-m1:
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -27,7 +27,7 @@ jobs:
     needs: prebuild
     with:
       name: Mac Intel
-      macos: macos-12
+      macos: macos-13
       kernel-target: x86_64-unknown-none
       artifact-name: obliteration-mac-intel
   build-mac-m1:


VocalFan:
Not sure if this is really a ME thing but...

```
error[E0531]: cannot find unit struct, unit variant or constant `VMX_REASON_HLT` in crate `hv_sys`
  --> gui/src/vmm/hv/macos/x86_64.rs:26:24
   |
26 |             Ok(hv_sys::VMX_REASON_HLT) => Ok(()),
   |                        ^^^^^^^^^^^^^^ not found in `hv_sys`
error[E0425]: cannot find value `HV_DEADLINE_FOREVER` in crate `hv_sys`
   --> gui/src/vmm/hv/macos/cpu.rs:149:71
    |
149 |             unsafe { hv_sys::hv_vcpu_run_until(self.instance, hv_sys::HV_DEADLINE_FOREVER) },
    |                                                                       ^^^^^^^^^^^^^^^^^^^ not found in `hv_sys`
error[E0425]: cannot find value `VMCS_RO_EXIT_REASON` in crate `hv_sys`
   --> gui/src/vmm/hv/macos/cpu.rs:159:29
    |
159 |                     hv_sys::VMCS_RO_EXIT_REASON,
    |                             ^^^^^^^^^^^^^^^^^^^ not found in `hv_sys`
    ```

ultimaweapon:
Strange. AFAIK `hv-sys` embeded a generated binding instead of generate it from the computer it is running. Switching macOS version should not cause this error.

VocalFan:
Yeah, and I'd rather our builds not get halted by "brownout" periods to make developers move to a newer version than macos-12

ultimaweapon:
@SuchAFuriousDeath do you want to maintain this? If not I think we should remove macOS x86 because no one want to maintenance it. It is going to die anyway since Apple already moved to ARM.

ultimaweapon:
Let's remove macOS x86.

SuchAFuriousDeath:
I agree.

PR #1031, opened 2024-10-12
Switches from hv-sys to applevisor-sys

ultimaweapon:
The new one looks better because it is hand written. It also have definition for `HV_MEMORY_*`.

This also remove CI for macOS x86.


VocalFan:
Screw OG Apple. ARM Apple is my *tolerable* friend.

SuchAFuriousDeath (approved):

PR #1048, opened 2024-10-19
Implements xmm and fpu internal reg getters

SuchAFuriousDeath:
I'm not sure about u128::from_le_bytes and how to name the getters for the internal fpu registers, the names in the Linux struct don't match the ones in the gdb_stub.

After adding mxcsr, we panic on read_registers.

### DIFF
--- a/gui/src/vmm/hv/linux/x86_64.rs
+++ b/gui/src/vmm/hv/linux/x86_64.rs
@@ -276,19 +276,19 @@ impl<'a> CpuStates for KvmStates<'a> {
     }
 
     fn get_fiseg(&mut self) -> Result<u32, Self::Err> {
-        Ok((self.fregs.last_ip & 0xFFFFFFFF) as u32)
-    }
-
-    fn get_fioff(&mut self) -> Result<u32, Self::Err> {
         Ok(((self.fregs.last_ip >> 32) & 0xFFFF) as u32)
     }
 
+    fn get_fioff(&mut self) -> Result<u32, Self::Err> {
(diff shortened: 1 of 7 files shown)


gui/src/vmm/hv/linux/x86_64.rs
```diff
@@ -262,6 +262,106 @@ impl<'a> CpuStates for KvmStates<'a> {
...
+        Ok((self.fregs.last_ip & 0xFFFFFFFF) as u32)
+    }
+
+    fn get_foseg(&mut self) -> Result<u32, Self::Err> {
+        Ok((self.fregs.last_dp >> 32) as u32)
+    }
+
+    fn get_fooff(&mut self) -> Result<u32, Self::Err> {
+        Ok((self.fregs.last_dp & 0xFFFFFFFF) as u32)
+    }
```
ultimaweapon, line 292: This ones looks weird. Should `seg` part come from the lower words instead of upper words?

SuchAFuriousDeath: Swapped

ultimaweapon: I'm not sure if my guess is correct. If you know your original version is a correct one then we should go with it.

ultimaweapon (approved):

gui/src/vmm/hv/linux/x86_64.rs
SuchAFuriousDeath: image
image

PR #1050, opened 2024-10-19
Implements read_registers for x86_64

SuchAFuriousDeath:


gui/src/vmm/cpu/x86_64.rs
```diff
@@ -97,4 +124,7 @@ impl<H: Hypervisor, S: Screen> ThreadExtraInfo for CpuManager<H, S> {
 
 /// Implementation of [`gdbstub::target::Target::Error`] for x86-64.
 #[derive(Debug, Error)]
-pub enum TargetError {}
+pub enum TargetError {
+    #[error("cpu not found for tid: {0}")]
+    CpuNotFound(Tid),
+}
```
ultimaweapon, line 130: From the documents this error represents a fatal error. That mean we should not use this for this case. We probably need to use `TargetError::Errno` with `ENOENT` instead.

ultimaweapon (changes requested):

gui/src/vmm/cpu/x86_64.rs
SuchAFuriousDeath: Done.

ultimaweapon (approved):

PR #1068, opened 2024-10-25
Implements MultiThreadResume for CpuManager

SuchAFuriousDeath:
I assume we want this.

### DIFF
--- a/gui/src/vmm/hv/linux/mod.rs
+++ b/gui/src/vmm/hv/linux/mod.rs
@@ -26,7 +26,7 @@ mod run;
 
 pub fn new(cpu: usize, ram: Ram, debug: bool) -> Result<Kvm, KvmError> {
     // Open KVM device.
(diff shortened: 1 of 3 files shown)


gui/src/vmm/hv/linux/mod.rs
```diff
@@ -26,7 +26,7 @@ mod run;
 
 pub fn new(cpu: usize, ram: Ram, debug: bool) -> Result<Kvm, KvmError> {
     // Open KVM device.
-    let kvm = unsafe { open(b"/dev/kvm\0".as_ptr().cast(), O_RDWR) };
+    let kvm = unsafe { open(c"/dev/kvm".as_ptr(), O_RDWR) };
```
ultimaweapon, line 29: cbindgen does not support this literal yet. Right now it does not break because this file is not processed by cbindgen due to`cfg` condition.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #1069, opened 2024-10-26
Implements software breakpoints for x86_64

SuchAFuriousDeath:


SuchAFuriousDeath:
One thing I'm not sure about is translating the address. I just use the 0 cpu

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
 
 impl<H: Hypervisor, S: Screen> SwBreakpoint for CpuManager<H, S> {
     fn add_sw_breakpoint(&mut self, addr: u64, kind: usize) -> TargetResult<bool, Self> {
-        todo!()
+        if self.sw_breakpoints.contains_key(&addr) {
+            return Ok(false);
+        }
+
+        let breakpoint_size = NonZero::new(kind).unwrap();
```
ultimaweapon, line 40: Do we actually need this? x86 only have one kind of breakpoint.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+
+        let cpu = self
+            .cpus
+            .first_mut()
+            .ok_or(TargetError::Errno(Self::GDB_ENOENT))?;
```
ultimaweapon, line 45: This should be a panic.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+
+        let translated_addr = cpu
+            .debug_mut()
+            .unwrap()
+            .translate_address(addr.try_into().unwrap())
+            .ok_or(TargetError::Errno(Self::GDB_ENOENT))?;
```
ultimaweapon, line 51: This should be something else. If the main CPU exit it probably a fatal error.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+
+        // Get data.
+        let src = self
+            .hv
+            .ram()
+            .lock(translated_addr, breakpoint_size)
+            .ok_or(TargetError::Errno(Self::GDB_EFAULT))?
+            .as_mut_ptr();
```
ultimaweapon, line 59: The `LockedAddr` is dropped at the end of statement. That mean the returned pointer will be invalid.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
 
 impl<H: Hypervisor, S: Screen> SwBreakpoint for CpuManager<H, S> {
     fn add_sw_breakpoint(&mut self, addr: u64, kind: usize) -> TargetResult<bool, Self> {
-        todo!()
+        if self.sw_breakpoints.contains_key(&addr) {
```
ultimaweapon, line 36: Better to use `entry` method for both check and insert.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+        Ok(true)
     }
 
     fn remove_sw_breakpoint(&mut self, addr: u64, kind: usize) -> TargetResult<bool, Self> {
-        todo!()
+        let Some(breakpoint) = self.sw_breakpoints.remove(&addr) else {
+            return Ok(false);
+        };
+
+        let breakpoint_size = NonZero::new(kind).unwrap();
```
ultimaweapon, line 80: Same as previous question.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+
+        if breakpoint.len() != breakpoint_size.get() {
+            todo!();
+        }
+
+        let cpu = self
+            .cpus
+            .first_mut()
+            .ok_or(TargetError::Errno(Self::GDB_ENOENT))?;
```
ultimaweapon, line 89: Same as previous issue.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+
+        let translated_addr = cpu
+            .debug_mut()
+            .unwrap()
+            .translate_address(addr.try_into().unwrap())
+            .ok_or(TargetError::Errno(Self::GDB_ENOENT))?;
```
ultimaweapon, line 95: Same here.

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -32,10 +33,79 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+
+        // Get data.
+        let src = self
+            .hv
+            .ram()
+            .lock(translated_addr, breakpoint_size)
+            .ok_or(TargetError::Errno(Self::GDB_EFAULT))?
+            .as_mut_ptr();
```
ultimaweapon, line 103: Same as previous issue.

ultimaweapon (changes requested):

gui/src/vmm/cpu/x86_64.rs
```diff
@@ -31,11 +36,60 @@ impl<H: Hypervisor, S: Screen> Breakpoints for CpuManager<H, S> {
...
+        // Get data.
+        let mut src = self
+            .hv
+            .ram()
+            .lock(translated_addr, BREAKPOINT_SIZE)
+            .ok_or(TargetError::Errno(Self::GDB_EFAULT))?;
+
+        let code_slice = src.as_mut_ptr().cast();
+
+        let mut code_bytes = std::mem::replace(&mut unsafe { *code_slice }, BREAKPOINT_BYTES);
```
ultimaweapon, line 61: I'm feeling unsafe here since `BREAKPOINT_SIZE` and `BREAKPOINT_BYTES` is a separated variable. Better to remove the whole `BREAKPOINT_BYTES` and hard-coded length of data to lock to `1`. Then use `std::mem::replace` on a single `u8` instead of an array. Then we can use `entry.insert([replaced])`, which will be automatically get type-check if someone change `BREAKPOINT_SIZE` to something else.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #1073, opened 2024-10-27
Implements MultiThreadResume for CpuManager

SuchAFuriousDeath:


SuchAFuriousDeath:
There are likely some things wrong in this PR, it requires review

ultimaweapon:
I'm not sure how `gdbstub` tell its debug target to resume. Does the target should resume when it reached `GdbStubStateMachine::Running` or it will be a job of `MultiThreadResume` implementation?

SuchAFuriousDeath:
I think it's the job of the resume method of MultiThreadResume

ultimaweapon:
In that case why `MultiThreadResume` is optional to implement? Is not the resuming after hitting a breakpoint should be a fundamental debug API?

SuchAFuriousDeath:
I asked AI and it said that it's optional, because in some cases, it's not possible to resume the target (such as when inspecting a core dump)

gui/src/vmm/cpu/mod.rs
```diff
@@ -520,19 +534,54 @@ impl<H: Hypervisor, S: Screen> ThreadExtraInfo for CpuManager<H, S> {
...
+
+        cpu.resume_action = Some(ResumeAction::Continue);
+
+        Ok(())
+    }
+
+    #[inline(always)]
+    fn support_single_step(&mut self) -> Option<MultiThreadSingleStepOps<'_, Self>> {
+        Some(self)
+    }
```
ultimaweapon, line 568: This should requires additional works on the hypervisor (e.g. `KVM_GUESTDBG_SINGLESTEP` on KVM). I'm not sure if the other hypervisor support this. If not we cannot enable this.

ultimaweapon (changes requested):

gui/src/vmm/cpu/mod.rs
SuchAFuriousDeath: I'll remove it for now.

ultimaweapon:
> I asked AI and it said that it's optional, because in some cases, it's not possible to resume the target (such as when inspecting a core dump)

In that case we need to remove the resuming code when we handled `GdbStubStateMachine::Running`.

SuchAFuriousDeath:
Not sure what exactly you mean. Is it this?

image

ultimaweapon:
Yes, we need to remove the call to `release` here.

SuchAFuriousDeath:
> Yes, we need to remove the call to `release` here.

Done.

gui/src/vmm/cpu/controller.rs
```diff
@@ -30,3 +32,7 @@ impl Drop for CpuController {
         unsafe { ManuallyDrop::take(&mut self.thread).join().unwrap() };
     }
 }
+
+pub(super) enum ResumeAction {
+    Continue,
+}
```
ultimaweapon, line 38: I don't think we need this. According to the docs of `MultiThreadResume::resume`:

```rust
    /// Any thread that wasn't explicitly resumed by a `set_resume_action_XXX`
    /// method should be resumed as though it was resumed with
    /// `set_resume_action_continue`.
```

That mean we don't even to do anything in `clear_resume_actions` and `set_resume_action_continue`.

ultimaweapon (approved):

PR #1084, opened 2024-11-03
Initializes Slint gui

### LINKED ISSUE #238: Try Slint as a new GUI
Now they just released a new version with a new license that is compatible with MIT license. So it is a good opportunity to try it to see if we can get rid of C++.

The new GUI should be placed in `src/gui` while we are experimenting with it. It should not ship with the release to the end users until it comes out from experimental.

SuchAFuriousDeath:
Closes #238.


ultimaweapon:
Require #1083 to merge first.

ultimaweapon:
You need to change `add_crate(gui)` to `add_crate(gui LIBRARY)`.

ultimaweapon:
#238 should be able to close by this PR.

ultimaweapon:
Seems like this PR is going too big. My idea is just pushing only the GUI without functionality first.

gui/build.rs
```diff
@@ -8,6 +9,30 @@ const LINUX_INCLUDE: &str = r#"
...
+    )
+    .unwrap();
+
+    // Compile main
+    slint_build::compile_with_config(
+        "slint/main.slint",
+        CompilerConfiguration::new().with_style(String::from("fluent-dark")),
+    )
+    .unwrap();
+}
```
ultimaweapon, line 33: Is it possible to use a wildcard here? Also I think it is better to rename the folder to the other name (e.g. `ui`) so we don't need to use `::slint` to reference the Slint crate.

gui/build.rs
```diff
@@ -8,6 +9,30 @@ const LINUX_INCLUDE: &str = r#"
 "#;
 
 fn main() {
+    if std::env::var("CARGO_FEATURE_GUI_SLINT").is_ok() {
```
ultimaweapon, line 12: There should be a better way than this. Feature flag should be used to enable additional code instead of using like this.

ultimaweapon (changes requested):

gui/build.rs
SuchAFuriousDeath: I tried, but AFAIK:
1. there can be only one build script per crate.
2. There is no way to tell if we are compiling the library or the binary.

gui/build.rs
SuchAFuriousDeath: I renamed the module, but wildcards don't seems possible based on the the code: Maybe with some compile time magic

ultimaweapon: In that case we can recursive list all files in the directory.

gui/build.rs
ultimaweapon: Have you try `CARGO_BIN_NAME`?

SuchAFuriousDeath: oh :no_mouth:

SuchAFuriousDeath: I tried CARGO_BIN_PATH, but it's not set :(.

ultimaweapon: Does Slint support including those files in Rust instead of building it here? Something like `include_slint!("file.slint").

SuchAFuriousDeath: I don't think so. The only other option is including slint source directly in Rust, there's a macro for that

ultimaweapon: Another idea is putting the FFI for Qt behind a feature flag and enabling it from CMake.

SuchAFuriousDeath: I did that, take a look

gui/build.rs
SuchAFuriousDeath: I removed the other file. It seems you can have only one slint file (You can export multiple things, but only from one file)

gui/build.rs
```diff
@@ -8,6 +9,23 @@ const LINUX_INCLUDE: &str = r#"
 "#;
 
 fn main() {
+    if std::env::var("CARGO_FEATURE_GUI_SLINT").is_ok_and(|var| var == "1") {
+        build_bin();
+    } else if std::env::var("CARGO_FEATURE_QT_FFI").is_ok_and(|var| var == "1") {
```
ultimaweapon, line 14: This should not be `else if`. See https://doc.rust-lang.org/cargo/reference/features.html#feature-unification for more information about this.

gui/build.rs
```diff
@@ -8,6 +9,23 @@ const LINUX_INCLUDE: &str = r#"
...
+    }
+}
+
+fn build_bin() {
+    // Compile main
+    slint_build::compile_with_config(
+        "slint/main.slint",
+        CompilerConfiguration::new().with_style(String::from("fluent-dark")),
+    )
+    .unwrap();
```
ultimaweapon, line 25: Better to recursive enumerate all files in this directory and compile all of `.slint` files.

gui/Cargo.toml
```diff
@@ -5,6 +5,19 @@ edition = "2021"
...
 crate-type = ["staticlib"]
+required-features = ["qt_ffi"]
+
+[[bin]]
+name = "gui"
+path = "src/main.rs"
+required-features = ["gui_slint"]
+
+[features]
+gui_slint = [
```
ultimaweapon, line 16: I think the name is too long. Just `slint` should be enough.

gui/Cargo.toml
```diff
@@ -5,6 +5,19 @@ edition = "2021"
...
+    "slint",
+    "clap"
+]
+qt_ffi = []
```
ultimaweapon, line 20: Same here, just `qt` should be enough.

gui/Cargo.toml
```diff
@@ -5,6 +5,19 @@ edition = "2021"
 
 [lib]
 crate-type = ["staticlib"]
+required-features = ["qt_ffi"]
+
+[[bin]]
+name = "gui"
```
ultimaweapon, line 11: This will be the name of output file so maybe `obliteration` should be better?

gui/Cargo.toml
```diff
@@ -23,9 +36,18 @@ uuid = { version = "1.10.0", features = ["serde", "v4"] }
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 x86-64 = { path = "../arch/x86-64" }
+raw-window-handle = "0.6"
```
ultimaweapon, line 39: Is not this should be optional?

gui/Cargo.toml
```diff
@@ -23,9 +36,18 @@ uuid = { version = "1.10.0", features = ["serde", "v4"] }
...
+clap = { version = "4.5.20", features = ["derive"], optional = true }
+
+[dependencies.slint]
+git = "https://github.com/slint-ui/slint"
+rev = "875ca075fb5b2dfe4c3ab0a499d5759412fc1395"
```
ultimaweapon, line 44: Is it possible to use the published version here? The same reason as `cbindgen`.

gui/Cargo.toml
```diff
@@ -46,3 +68,4 @@ objc = "0.2.7"
 
 [build-dependencies]
 cbindgen = "0.26.0"
+slint-build = { git = "https://github.com/slint-ui/slint", rev = "875ca075fb5b2dfe4c3ab0a499d5759412fc1395" }
```
ultimaweapon, line 71: Same here.

cmake/cargo.cmake
```diff
@@ -157,6 +157,10 @@ function(add_crate crate)
     list(APPEND build_args "$<IF:$<CONFIG:Debug>,--profile=dev,--release>")
     list(APPEND build_args ${arg_ARGS})
 
+    if(arg_LIBRARY)
+        list(APPEND build_args "--lib")
+    endif()
+
```
ultimaweapon, line 163: This one should be inside `staticlib` block below so it works with a single type crate.

ultimaweapon (changes requested):

gui/build.rs
SuchAFuriousDeath: We will only be compiling one file. Slint doesn't support multiple files.

gui/Cargo.toml
SuchAFuriousDeath: Not at the moment

gui/Cargo.toml
SuchAFuriousDeath: Same here too :(

gui/Cargo.toml
ultimaweapon: In that case we should create an issue for this so when the release we need available we can fix this.

cmake/cargo.cmake
```diff
@@ -178,6 +178,11 @@ function(add_crate crate)
 
         if(${kind} STREQUAL "staticlib")
             add_library(${crate} STATIC IMPORTED)
+
+            if(arg_LIBRARY)
```
ultimaweapon, line 182: We don't need this `if` since we are in the `staticlib` block.

gui/src/screen/vulkan/mod.rs
```diff
@@ -31,77 +19,44 @@ pub struct Vulkan {
...
+        let appinfo = ApplicationInfo::default()
+            .application_name(c"Obliteration")
+            .application_version(0)
+            .api_version(ash::vk::make_api_version(0, 1, 0, 0));
+
+        let create_info = InstanceCreateInfo::default().application_info(&appinfo);
+
+        let entry = ash::Entry::linked();
+
+        let instance = unsafe { entry.create_instance(&create_info, None) }
```
ultimaweapon, line 33: This should reuse the same instance from the Qt side.

ultimaweapon (changes requested):

VocalFan:
I will just keep watching from the distance... Also known as emails from github.

gui/src/rlim.rs
```diff
@@ -0,0 +1,32 @@
...
+                    if unsafe { libc::setrlimit(libc::RLIMIT_NOFILE, &rlim) } != 0 {
+                        eprintln!(
+                            "Failed to set RLIMIT_NOFILE: {}",
+                            std::io::Error::last_os_error()
+                        );
+                    }
+                }
+            }
+            _ => {
+                eprintln!(
```
ultimaweapon, line 25: This should be a message box otherwise the user won't know what is going wrong.

gui/src/rlim.rs
```diff
@@ -0,0 +1,32 @@
...
+
+        match ret {
+            0 => {
+                let mut rlim = unsafe { rlim.assume_init() };
+
+                if rlim.rlim_cur < rlim.rlim_max {
+                    rlim.rlim_cur = rlim.rlim_max;
+
+                    if unsafe { libc::setrlimit(libc::RLIMIT_NOFILE, &rlim) } != 0 {
+                        eprintln!(
```
ultimaweapon, line 17: Same here.

ultimaweapon (changes requested):

gui/src/rlim.rs
```diff
@@ -0,0 +1,34 @@
...
+
+            Ok(())
+        }
+        _ => Err(RlimitError::GetRlimitFailed(std::io::Error::last_os_error())),
+    }
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum RlimitError {
+    #[error("failed to get RLIMIT_NOFILE -> {0}")]
```
ultimaweapon, line 29: Printing the nested error here will make it impossible for the caller to format the error chain. Better to use `Error::source` on the caller side instead.

gui/src/rlim.rs
```diff
@@ -0,0 +1,34 @@
...
+    GetRlimitFailed(#[source] std::io::Error),
+
+    #[error("failed to set RLIMIT_NOFILE -> {0}")]
```
ultimaweapon, line 32: Same here.

ultimaweapon (changes requested):

gui/src/vmm/hv/mod.rs
```diff
@@ -24,6 +24,15 @@ pub type Default = self::os::Hvf;
...
 
+#[cfg(target_os = "linux")]
+pub type DefaultError = self::os::KvmError;
+
+#[cfg(target_os = "macos")]
+pub type DefaultError = self::os::HvfError;
+
+#[cfg(target_os = "windows")]
+pub type DefaultError = self::os::WhpError;
+
```
ultimaweapon, line 35: These should be `HypervisorError`.

ultimaweapon (changes requested):

gui/src/vmm/mod.rs
```diff
@@ -684,6 +444,116 @@ pub enum DebugResult {
     Error { reason: *mut RustError },
 }
 
+#[derive(Debug, Error)]
+pub enum StartVmmError {
```
ultimaweapon, line 448: Now `VmmError` has been removed so we can reuse this name here.

ultimaweapon (changes requested):

gui/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+    Ok,
+    Err(ApplicationError),
+}
+
+impl Termination for AppExit {
+    fn report(self) -> ExitCode {
+        match self {
+            AppExit::Ok => ExitCode::SUCCESS,
+            AppExit::Err(e) => {
+                eprintln!("{}", e);
```
ultimaweapon, line 130: We should use a message box instead otherwise it won't be seen by the user.

gui/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl Termination for AppExit {
+    fn report(self) -> ExitCode {
+        match self {
+            AppExit::Ok => ExitCode::SUCCESS,
+            AppExit::Err(e) => {
+                let _ = ui::ErrorDialog::new()
+                    .and_then(|error_dialog| {
+                        error_dialog.set_message(SharedString::from(format!(
+                            "Error setting rlimit: {}",
```
ultimaweapon, line 133: Seems like the wrong message here.

gui/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+                            full_error_reason(e)
+                        )));
+
+                        error_dialog.run()
+                    })
+                    .inspect_err(|e| eprintln!("Error displaying error dialog: {e}"));
```
ultimaweapon, line 139: I think unwrap should be better because this should never fail.

gui/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+        let _ = ui::ErrorDialog::new()
+            .and_then(|error_dialog| {
+                error_dialog.set_message(SharedString::from(format!(
+                    "Error setting rlimit: {}",
+                    full_error_reason(e)
+                )));
+
+                error_dialog.run()
+            })
+            .inspect_err(|e| eprintln!("Error displaying error dialog: {e}"));
```
ultimaweapon, line 40: Same here.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #1093, opened 2024-11-11
Passes device list to Slint UI

SuchAFuriousDeath:


gui/src/screen/mod.rs
```diff
@@ -7,17 +7,37 @@ use std::sync::Arc;
...
+
+    fn init() -> Result<Self, Self::InitError>;
+
+    fn enumerate_physical_devices(&self) -> &[Self::PhysicalDevice];
+}
+
+pub trait PhysicalDevice: Sized {
+    fn name(&self) -> &str;
+}
+
```
ultimaweapon, line 40: This and its implementations should be a top level module.

ultimaweapon (changes requested):

gui/src/screen/mod.rs
SuchAFuriousDeath: Like src/graphics?

ultimaweapon: Right.

ultimaweapon:
Wait, I mean top-level module of the `gui` crate, not a separated crate.

gui/src/graphics/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(target_os = "macos")]
+pub type DefaultApi = self::api::Metal;
+
+pub trait GraphicsApi: Sized + 'static {
+    type PhysicalDevice: PhysicalDevice;
+
+    type InitError: core::error::Error;
+
+    fn init() -> Result<Self, Self::InitError>;
```
ultimaweapon, line 18: I think this name cause confusion. When I see this name what I understand is it a module-level initialization like https://docs.rs/log/latest/log/fn.set_logger.html. So better to use standard Rust naming like `new` instead.

gui/src/graphics/metal.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl super::PhysicalDevice for metal::Device {
+    fn name(&self) -> &str {
+        self.name()
+    }
+}
+
+/// Represents an error when [`Metal::init()`] fails.
+#[derive(Debug, Error)]
+pub enum MetalInitError {}
```
ultimaweapon, line 40: Once `GraphicsApi::init` has been renamed this need to be renamed too.

gui/src/graphics/vulkan.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl super::PhysicalDevice for VulkanPhysicalDevice {
+    fn name(&self) -> &str {
+        &self.name
+    }
+}
+
+/// Represents an error when [`Vulkan::init()`] fails.
+#[derive(Debug, Error)]
+pub enum VulkanInitError {
```
ultimaweapon, line 76: Same here.

ultimaweapon (changes requested):

gui/src/graphics/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub type DefaultApi = self::api::Metal;
+
+pub trait GraphicsApi: Sized + 'static {
+    type PhysicalDevice: PhysicalDevice;
+
+    type CreateError: core::error::Error;
+
+    fn new() -> Result<Self, Self::CreateError>;
+
+    fn enumerate_physical_devices(&self) -> &[Self::PhysicalDevice];
```
ultimaweapon, line 20: The method name tell it is going to enumerate the devices while the actual behavior is returned a list of devices that already enumerated so better to rename to just `physical_devices`.

SuchAFuriousDeath: good point

gui/src/graphics/vulkan.rs
```diff
@@ -0,0 +1,90 @@
...
+            .map_err(VulkanCreateError::CreateInstanceFailed)?;
+
+        let devices = unsafe { instance.enumerate_physical_devices() }
+            .map_err(VulkanCreateError::EnumeratePhysicalDevicesFailed)?
+            .into_iter()
+            .map(
+                |device| -> Result<VulkanPhysicalDevice, VulkanCreateError> {
+                    let properties = unsafe { instance.get_physical_device_properties(device) };
+
+                    let name = CStr::from_bytes_until_nul(unsafe {
```
ultimaweapon, line 34: Why don't use `CStr::from_ptr` instead of making an intermediate slice?

ultimaweapon (changes requested):

gui/src/graphics/vulkan.rs
SuchAFuriousDeath: In theory, there could be no null bytes in the 256 bytes. This way, we can catch that.

ultimaweapon: In that case better to remove a hard-coded length otherwise it will be UB if the length does not match in the future.

ultimaweapon (approved):

PR #1103, opened 2024-11-16
Removes raw_window_handle, uses Cstr literals

SuchAFuriousDeath:


ultimaweapon:
Is not we need `raw_window_handle` to access Vulkan surface and Metal layer?

SuchAFuriousDeath:
> I though we need `raw_window_handle` to access Vulkan surface and Metal layer?
Eventually, but we do not depende on it right now. I'll put it back once it is actually required.

ultimaweapon (approved):

PR #1108, opened 2024-11-16
Initialiazes Wizard

SuchAFuriousDeath:
I'm really falling in love with Slint


Cargo.lock
```diff
@@ -42,7 +42,7 @@ dependencies = [
  "atspi-common",
  "serde",
  "thiserror",
- "zvariant",
+ "zvariant 4.2.0",
```
ultimaweapon, line 45: Is it possible to use a single version for this crate?

Cargo.lock
```diff
@@ -84,7 +84,7 @@ dependencies = [
  "futures-lite",
  "futures-util",
  "serde",
- "zbus",
+ "zbus 4.4.0",
```
ultimaweapon, line 87: Same here.

ultimaweapon (changes requested):

Cargo.lock
SuchAFuriousDeath: Doesn't look like it

SuchAFuriousDeath: Actually, yes.

Cargo.lock
SuchAFuriousDeath: Solved

SuchAFuriousDeath:
When winit gets a new release, we could get rid of the duplicate bitflags

gui/Cargo.toml
```diff
@@ -34,6 +34,7 @@ ciborium = "0.2.2"
 serde = { version = "1.0.209", features = ["derive"] }
 thiserror = "1.0"
 uuid = { version = "1.10.0", features = ["serde", "v4"] }
+rfd = "0.14.0"
```
ultimaweapon, line 37: This should be optional.

ultimaweapon (changes requested):

gui/slint/main/tabs/cpu.slint
```diff
@@ -7,18 +7,31 @@ export component CpuTab {
...
+
             Text {
                 text: "Count";
             }
-            Slider {
-                value: 8;
-                minimum: 1;
-                maximum: 16;
+            HorizontalLayout {
+                spacing: 5px;
```
ultimaweapon, line 16: Better to use `HorizontalBox` instead of specify spacing manually.

gui/slint/main/tabs/cpu.slint
```diff
@@ -7,18 +7,31 @@ export component CpuTab {
...
+                    text: cpus;
+                }
             }
             Text {
                 text: "Changing this value to other than 8 may crash the game.";
             }
-        }  
+        }
         VerticalLayout {
             padding: 10px;
```
ultimaweapon, line 36: Same here.

gui/src/main.rs
```diff
@@ -22,12 +23,26 @@ mod ui;
...
+        ui::ErrorDialog::new()
+            .and_then(|error_dialog| {
+                error_dialog.set_message(SharedString::from(format!(
+                    "Error running application: {}",
+                    full_error_reason(e)
+                )));
+
+                error_dialog.run()
+            })
+            .inspect_err(|e| eprintln!("Error displaying error dialog: {e}"))
```
ultimaweapon, line 36: This line should not be necessary since unwrap already printing the debug.

gui/src/main.rs
```diff
@@ -43,68 +58,156 @@ fn run() -> Result<(), ApplicationError> {
...
+        VecModel::from(vec![profile::Profile::default()])
+            .map(|p| SharedString::from(String::from(p.name().to_string_lossy()))),
+    );
+
+    main_window.set_profiles(profiles.clone());
+
+    main_window.on_start_game(|_index| {
+        let screen = ui::Screen::new().unwrap();
+
+        screen.show().unwrap();
```
ultimaweapon, line 101: Is it possible to reuse the same Windows like how Qt does?

ultimaweapon (changes requested):

gui/src/main.rs
SuchAFuriousDeath: It might be. Let's put a todo there for since there's a lot of work to be done there anyway

gui/slint/main/tabs/cpu.slint
ultimaweapon: This one still `VerticalLayout`.

gui/slint/wizard.slint
```diff
@@ -0,0 +1,248 @@
...
+    NotFile,
+    SameAsSystemDir
+}
+
+export component Wizard inherits Window {
+    min-width: 1000px;
+    min-height: 500px;
+
+    pure callback validate_system_dir(string) -> FileValidationResult;
+    pure callback validate_games_dir(string, string) -> FileValidationResult;
```
ultimaweapon, line 27: We should not need this since the game management will be handled by the PS4 system software.

ultimaweapon (changes requested):

gui/slint/wizard.slint
ultimaweapon: Wait, I guess we need something like this. The idea is the PS4 should try to mount some RW partitions as a storage to install the game and other data. But let's remove this for now until we know exactly how many RW partition the PS4 will try to mount.

SuchAFuriousDeath: Just removed it.

SuchAFuriousDeath:
I opened this Slint issue btw, it should be possible to unify ttf-parser to a single version (we currently compile it three times)
https://github.com/slint-ui/slint/issues/6806

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+    min-width: 1000px;
+    min-height: 500px;
+
+    pure callback validate_firmware_path(string) -> FileValidationResult;
+
+    pure callback cancel();
+
+    property <WizardPage> page: WizardPage.Intro;
+    property <string> system_dir;
+    property <string> games_dir;
```
ultimaweapon, line 28: No once used these to better to remove it.

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+    property <string> firmware_path;
+
+    VerticalLayout {
+        if page == WizardPage.Intro: VerticalLayout {
+            padding: 10px;
```
ultimaweapon, line 33: Better to use `VerticalBox` instead.

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+                }
+                Button {
+                    text: "Cancel";
+                    clicked => { cancel(); }
+                }
+            }
+        }
+
+        if page == WizardPage.Firmware: VerticalLayout {
+            padding: 10px;
```
ultimaweapon, line 65: Same here.

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+                }
+                Button {
+                    text: "Cancel";
+                    clicked => { cancel(); }
+                }
+            }
+        }
+
+        if page == WizardPage.Conclusion: VerticalLayout {
+            padding: 10px;
```
ultimaweapon, line 117: Same here.

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+                Text {
+                    text: "Install Firmware";
+                    font-weight: 700;
+                }
+                Text {
+                    text: "Select a firmware dump that you got from Firmware Dumper.";
+                    vertical-stretch: 1;
+                }
+                HorizontalLayout {
+                    spacing: 5px;
```
ultimaweapon, line 81: Same here.

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+        if page == WizardPage.Conclusion: VerticalLayout {
+            padding: 10px;
+            vertical-stretch: 1;
+
+            Text {
+                text: "Setup complete";
+                font-weight: 700;
+            }
+            Text {
+                text: "You can now install your games and play them using Obliteration.";
```
ultimaweapon, line 125: We should change the text here since the user will need to do things on the PS4 system software instead.

ultimaweapon (changes requested):

gui/slint/wizard.slint
```diff
@@ -0,0 +1,146 @@
...
+                HorizontalLayout {
+                    spacing: 5px;
+
+                    Text {
+                        text: "File:";
+                        vertical-alignment: center;
+                    }
+                    LineEdit {
+                        text <=> firmware_path;
+                        max-height: 30px;
```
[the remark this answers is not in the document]
SuchAFuriousDeath: We don't now, this is a leftover from when I was struggling with the layouts some time ago :sweat:

ultimaweapon (approved):

PR #1115, opened 2024-11-18
Removes unnecessary caches

ultimaweapon:
`.flatpak-builder` is very small so caching does not make sense. For Cargo home it is almost 1 GB, which seems too big. I believe it is because its contains every crates that we have used since we keep caching it. So I thing it should be better not to cache this. The Windows CI also spend a lot of time to restore this so remove this should reduce make it faster.


SuchAFuriousDeath (approved):

ultimaweapon:
Looking good. We can cut time from 12 minutes to 8 minutes on Windows!

PR #1116, opened 2024-11-19
Adds debug launch configuration and task for Slint

SuchAFuriousDeath:


.vscode/launch.json
```diff
@@ -40,13 +40,30 @@
...
       },
+      "args": [
+        "--kernel",
+        "${workspaceFolder}/build/obkrnl"
+      ],
       "cwd": "${workspaceFolder}"
     },
     {
-      "name": "Kernel",
+      "name": "Kernel (Qt)",
```
ultimaweapon, line 50: The parentheses here does not look good on the VS Code due to VS Code also put some text after this with parentheses. Better to change to another format (e.g. `Kernel | Qt`).

.vscode/launch.json
```diff
@@ -40,13 +40,30 @@
...
+      "targetCreateCommands": [
+        "target create ${workspaceFolder}/build/obkrnl",
+        "target modules load --file ${workspaceFolder}/build/obkrnl -s 0xffffffff82200000"
+      ],
+      "processCreateCommands": [
+        "gdb-remote 1234"
+      ]
+    },
+    {
+      "name": "Kernel (Slint)",
```
ultimaweapon, line 63: Same here.

ultimaweapon (changes requested):

gui/src/main.rs
```diff
@@ -183,21 +234,30 @@ enum ApplicationError {
     #[error("failed to run wizard")]
     RunWizard(#[source] slint::PlatformError),
 
+    #[error("get current executable path")]
+    GetCurrentExe(#[source] std::io::Error),
+
```
ultimaweapon, line 239: This does not seems right.

ultimaweapon (changes requested):

gui/src/main.rs
SuchAFuriousDeath: What about it?

SuchAFuriousDeath: Done

ultimaweapon (approved):

PR #1118, opened 2024-11-19
Refactors event handler

SuchAFuriousDeath:
I guess VmmEventHandle will have to wrap an impl FnMut(VmmEvent) + Clone closure? Or something like that.

### DIFF
--- a/gui/src/main.rs
+++ b/gui/src/main.rs
@@ -1,10 +1,11 @@
+use self::profile::Profile;
 use self::ui::ErrorDialog;
 use self::vmm::Vmm;
 use args::CliArgs;
 use clap::Parser;
 use debug::DebugServer;
 use graphics::{GraphicsApi, PhysicalDevice};
-use slint::{ComponentHandle, Global, ModelExt, ModelRc, SharedString, VecModel};
+use slint::{ComponentHandle, Global, ModelRc, SharedString, VecModel};
 use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 use thiserror::Error;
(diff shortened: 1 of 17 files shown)


ultimaweapon:
> I guess VmmEventHandle will have to wrap an impl FnMut(VmmEvent) + Clone closure?

Should be `Arc<dyn Fn(...)>` instead.

SuchAFuriousDeath:
> > I guess VmmEventHandle will have to wrap an impl FnMut(VmmEvent) + Clone closure?
> 
> Should be `Arc<dyn Fn(...)>` instead.

Close enough :sweat_smile:

ultimaweapon:
I don't think global models are a good idea. Why do we need it? It should be passing to the main window and downstream to all of its children instead.

SuchAFuriousDeath:
> > I don't think global models are a good idea. Why do we need it? It should be passing to the main window and downstream to all of its children instead.
> 
> When launching in debug, there is no main window (and the screen isn't its child anyway). Maybe that's the fix that needs fixing here 🤔

But that's actually not the real problem. The real problem is that I don't like having to pass the devices and profiles through several elements down to where we actually need them. If they're global, we can just directly access them right where we need them (in the ComboBoxes)

ultimaweapon:
> > > I don't think global models are a good idea. Why do we need it? It should be passing to the main window and downstream to all of its children instead.
> > 
> > 
> > When launching in debug, there is no main window (and the screen isn't its child anyway). Maybe that's the fix that needs fixing here 🤔
> 
> But that's actually not the real problem. The real 'problem' is that I don't like having to pass the devices and profiles through several elements down to where we actually need them. If they're global, we can just directly access them right where we need them (in the ComboBoxes)

What you don't like here are actually a good practice.

SuchAFuriousDeath:
> > > > I don't think global models are a good idea. Why do we need it? It should be passing to the main window and downstream to all of its children instead.
> > > 
> > > 
> > > When launching in debug, there is no main window (and the screen isn't its child anyway). Maybe that's the fix that needs fixing here 🤔
> > 
> > 
> > But that's actually not the real problem. The real 'problem' is that I don't like having to pass the devices and profiles through several elements down to where we actually need them. If they're global, we can just directly access them right where we need them (in the ComboBoxes)
> 
> What you don't like here are actually a good practice.

Oops :D

gui/src/main.rs
```diff
@@ -78,24 +89,28 @@ fn run() -> Result<(), ApplicationError> {
...
-    setup_global_callbacks(&main_window);
-
     let graphics_api = graphics::DefaultApi::new().map_err(ApplicationError::InitGraphicsApi)?;
 
-    let devices: Vec<SharedString> = graphics_api
-        .physical_devices()
-        .into_iter()
-        .map(|d| SharedString::from(d.name()))
-        .collect();
+    let profiles = vec![Profile::default()];
```
ultimaweapon, line 94: Better to centralize this now so we don't have a duplicated code.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #1119, opened 2024-11-19
Revises Slint menu

ultimaweapon:
Screenshot From 2024-11-20 03-33-03


SuchAFuriousDeath:
:hot_face:

SuchAFuriousDeath (approved):

ultimaweapon:
Actually I never though we are able to have a good looking Rust GUI. The only problems right now is native menu but now it is okay since we have a workaround. Seems like we can ditch C++ and Qt sooner than expected.

PR #1131, opened 2024-11-22
Refactors vmm events

SuchAFuriousDeath:


gui/src/vmm/mod.rs
```diff
@@ -440,23 +440,12 @@ pub struct VmmScreen {
...
-        ty: VmmLog,
-        data: *const c_char,
-        len: usize,
-    },
-    Breakpoint {
-        stop: *mut KernelStop,
-    },
+    Error { reason: RustError },
+    Exiting { success: bool },
+    Log { ty: VmmLog, msg: Box<[u8]> },
```
ultimaweapon, line 447: Now we can use Rust string here since it no longer expose to C++ side.

gui/src/vmm/mod.rs
```diff
@@ -440,23 +440,12 @@ pub struct VmmScreen {
 pub type VmmEventHandler = Arc<dyn Fn(VmmEvent) + Send + Sync + 'static>;
 
 /// Contains VMM event information.
-#[repr(C)]
 #[allow(dead_code)] // TODO: Figure out why Rust think fields in each enum are not used.
```
ultimaweapon, line 443: This one no longer needed.

ultimaweapon (changes requested):

ultimaweapon (changes requested):

gui/src/hv/ram.rs
```diff
@@ -302,6 +302,12 @@ impl<'a> LockedAddr<'a> {
     }
 }
 
+impl AsRef<[u8]> for LockedAddr<'_> {
+    fn as_ref(&self) -> &[u8] {
+        unsafe { std::slice::from_raw_parts(self.ptr, self.len.get()) }
+    }
+}
+
```
ultimaweapon, line 310: We can't do this because the returned slice subject to the data race, which is UB for Rust.

gui/src/vmm/hw/console/context.rs
```diff
@@ -44,30 +44,29 @@ impl<'a, H: Hypervisor, C: Cpu> DeviceContext<C> for Context<'a, H> {
...
             // Trigger event.
             let msg = std::mem::take(&mut self.msg);
 
-            (self.dev.event)(VmmEvent::Log {
-                ty: ty.into(),
-                data: msg.as_ptr().cast(),
-                len: msg.len(),
-            });
+            // Safety: We check the message is valid UTF-8 above.
+            let msg = unsafe { String::from_utf8_unchecked(msg) };
```
ultimaweapon, line 67: Why not use `to_owned` on the returned value of `from_utf8` above?

ultimaweapon (changes requested):

gui/src/vmm/hw/console/context.rs
SuchAFuriousDeath: Because that would clone the whole vector when we don't have to.

SuchAFuriousDeath: I did it like this because it preserves the logic of the original code - check utf8 (and error if not utf8), check commit, mem::take the Vec, and then construct the VmmEvent infallibly.
String::from_utf8 takes and owned vec, meaning that we have to mem::take self.msg and then check if the string is utf8 and it it isn't, put the Vec back into self.msg. Or maybe you're fine with not putting it back. I don't know.
What exactly do you want?

ultimaweapon: Every errors here will exit the VMM immediately so we don't need to put it back.

gui/src/vmm/hw/console/context.rs
```diff
@@ -89,7 +85,7 @@ enum ExecError {
     InvalidLen,
 
     #[error("invalid message")]
-    InvalidMsg,
+    InvalidMsg(#[from] std::string::FromUtf8Error),
```
ultimaweapon, line 88: This can cause wrong behavior if someone accidentally `String::from_utf8(v)?` on a non-msg field.

gui/src/main.rs
```diff
@@ -81,22 +81,17 @@ fn run() -> Result<(), ApplicationError> {
...
         .map_err(ApplicationError::RunVmm)?;
-    }
 
-    // Run VMM launcher.
-    let vmm = match run_launcher()? {
-        Some(v) => v,
-        None => return Ok(()),
+        todo!()
+    } else {
+        run_launcher()?
```
ultimaweapon, line 87: `run_launcher` will return `Ok(None)` if the user exit the window instead of clicking on `Start` button. In this case we need to return immediately without creating the screen.

ultimaweapon (changes requested):

gui/src/vmm/hw/console/context.rs
SuchAFuriousDeath: Yeah, that was copilot. :sweat_smile:

gui/src/main.rs
SuchAFuriousDeath: We will return right in the next code

ultimaweapon (approved):

PR #1133, opened 2024-11-23
Prepares to merge VMM screen into graphics module

ultimaweapon:
It should be better for `GraphicsApi` implementation to create a VMM screen so it can share private types. The idea is the GUI will configure the `GraphicsApi` then the `GraphicsApi` will create a VMM screen according to the configurations.


SuchAFuriousDeath:
I envisioned it that way.

SuchAFuriousDeath (approved):

PR #1140, opened 2024-11-24
Replaces Flatpak with normal binary

ultimaweapon:
Now our binary only depend on `libc` and Vulkan so it should be able to run on any distros without Flatpak.


ultimaweapon:
@SuchAFuriousDeath can you test the binary on your distro to see if it work?

SuchAFuriousDeath:
It does

SuchAFuriousDeath (approved):

PR #1149, opened 2024-11-30
Prevents main window from running when wizard closes, appeases clippy

SuchAFuriousDeath:
I saw that one of the commits in a recent PR addressed a lot of lifetime clippy lints, so I fixed the rest


gui/src/main.rs
```diff
@@ -234,13 +234,12 @@ fn run_launcher(
     drop(win);
 
     // Extract GUI states.
-    let start = Rc::into_inner(start).unwrap().into_inner();
+    let start = Rc::into_inner(start)
+        .unwrap()
+        .into_inner()
+        .then_some(VmmArgs {});
```
ultimaweapon, line 240: This one should not construct here because we need to pull the other GUI states to construct it (e.g. debug address).

ultimaweapon (changes requested):

gui/src/main.rs
SuchAFuriousDeath: Ok, but we can refactor that later, this is equivalent to what's in the code now, just more 'Rusty'

SuchAFuriousDeath: On the other hand, this is a simple fi

ultimaweapon (approved):

PR #1151, opened 2024-11-30
Refactors vmm args

SuchAFuriousDeath:


gui/src/main.rs
```diff
@@ -147,8 +145,13 @@ fn run_vmm(args: &Args) -> Result<(), ApplicationError> {
 
         todo!()
     } else {
-        match run_launcher(&graphics, profiles)? {
-            Some(v) => v,
+        let vmm_args = run_launcher(&graphics, profiles)?.then_some(VmmArgs {
+            kernel_path,
+            debugger: None,
```
ultimaweapon, line 150: How we will construct this field here? I mean when the user click on the start debug button next to the debug address.

ultimaweapon (changes requested):

gui/src/main.rs
```diff
@@ -241,11 +252,11 @@ fn run_launcher(
 
     win.on_start_vmm({
         let win = win.as_weak();
-        let start = start.clone();
+        let debug_addr = debug_addr.clone();
 
         move || {
             win.unwrap().hide().unwrap();
-            start.set(true);
+            debug_addr.set(Some(None));
```
ultimaweapon, line 259: Does this mean we always start the kernel in debug when pressing Start button?

ultimaweapon (changes requested):

gui/src/main.rs
ultimaweapon: I think I understand it now. Better to create a dedicated enum for this since nested `Option` here is really confusing.

gui/src/main.rs
```diff
@@ -226,11 +238,11 @@ fn run_panic_handler() -> Result<(), ApplicationError> {
 fn run_launcher(
     graphics: &impl Graphics,
     profiles: Vec<Profile>,
-) -> Result<Option<VmmArgs>, ApplicationError> {
+) -> Result<Option<ExitAction>, ApplicationError> {
     // Create window and register callback handlers.
     let win = MainWindow::new().map_err(ApplicationError::CreateMainWindow)?;
     let profiles = Rc::new(ProfileModel::new(profiles));
-    let start = Rc::new(Cell::new(false));
+    let debug_addr = Rc::new(Cell::new(None));
```
ultimaweapon, line 245: `debug_addr` does not make sense here.

ultimaweapon (changes requested):

ultimaweapon (approved):

PR #1208, opened 2024-12-28
Removes gdbstub dispatcher

ultimaweapon:
Too hard and not ergonomic with a custom event loop. I though it was because of Qt but this problem still happens with our Rust event loop. This is a first step to remove it and we will implement our own one in the future.


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
Would it help if gdbstub had an async api? There'd been discussions of that.

ultimaweapon:
> Would it help if gdbstub had an async api? There'd been discussions of that.

Need to see the proposal first before I can tell. The main problem I encountered is caused by their state machine. It will be a lot easier if they have only one type to provide a method to feed the data received from the debugger and a stop reason from the debuggee something like:

```rust
impl GdbStub {
    pub fn data_received(&mut self, data: &[u8]);
    pub fn handle_stop(&mut self, reason: MultiThreadStopReason<u64>);
}
```

Their current design forced us to drive their state machine as a primary task and feed what it need on each state, which may not available yet. It should be inverse as my above idea.

SuchAFuriousDeath:
This is probably too deep for me 😅

SuchAFuriousDeath:
Create an issue in gdbstub, maybe something can be done about it.

ultimaweapon:
> Create an issue in gdbstub, maybe something can be done about it.

We probably the one who will implement it if we do that, the same as https://github.com/daniel5151/gdbstub/issues/153.

SuchAFuriousDeath:
Yup.

SuchAFuriousDeath:
Maybe we could get some suggestions and implement something

ultimaweapon:
I'm tired with it and don't want to work on this but if you can we can continue using it instead of implement our own one. If not I'll implement our own one in the future when I recovered from this suffer.

SuchAFuriousDeath:
Ok.

daniel5151:
Heya! I'm not sure I totally understand the limitation that you ran into here, but I would be very interested in learning more, and refining `gdbstub`'s API to be more amenable to your use-case. This wouldn't be the first time I've needed to do so - in fact, this entire state machine API that you just removed was the result of a collaborative effort between me and some folks who wanted to use `gdbstub` in the context of their `no_std` kernel / embedded projects!

If you feel up to it, feel free to file an issue on the `gdbstub` repo so we can discuss the pain points you ran into, and how `gdbstub` might be improved to make life easier for you / others in the future.

* * *

Sorry for popping in unannounced 😅 I've got a saved GitHub search for downstream mentions of `gdbstub` that I check once in a while, which is how I stumbled across this particular issue.

SuchAFuriousDeath:
Good thing I mentioned that issue I guess 😅

ultimaweapon:
Okay I'll create an issue to report the problem I found when I'm ready. Sorry for the rant.

daniel5151:
Hah, no worries

ultimaweapon:
Sorry to let you know that I can't remember all of the details about the problem I found. In order to recover those problems from my memory I need to make another attempt to integrate the `gdbsub` into our event loop but the main problem is what I posed on https://github.com/obhq/obliteration/pull/1208#issuecomment-2564301410. The following code was our successful integration with the Qt event loop:

```rust
    fn dispatch_debug(
        &mut self,
        mut stop: Option<MultiThreadStopReason<u64>>,
    ) -> Result<DispatchDebugResult, DispatchDebugError> {
        loop {
            // Check current state.
            let r = match self.gdb.take().unwrap() {
                GdbStubStateMachine::Idle(s) => {
                    match self.cpu.dispatch_gdb_idle(s) {
                        Ok(Ok(v)) => Ok(v),
                        Ok(Err(v)) => {
                            // No pending data from the debugger.
                            self.gdb = Some(v.into());
                            return Ok(DispatchDebugResult::Ok);
                        }
                        Err(e) => Err(DispatchDebugError::DispatchIdle(e)),
                    }
                }
                GdbStubStateMachine::Running(s) => {
                    match self.cpu.dispatch_gdb_running(s, stop.take()) {
                        Ok(Ok(v)) => Ok(v),
                        Ok(Err(v)) => {
                            // No pending data from the debugger.
                            self.gdb = Some(v.into());
                            return Ok(DispatchDebugResult::Ok);
                        }
                        Err(e) => Err(DispatchDebugError::DispatchRunning(e)),
                    }
                }
                GdbStubStateMachine::CtrlCInterrupt(s) => {
                    self.cpu.lock();

                    s.interrupt_handled(
                        &mut self.cpu,
                        Some(MultiThreadStopReason::Signal(Signal::SIGINT)),
                    )
                    .map_err(DispatchDebugError::HandleInterrupt)
                }
                GdbStubStateMachine::Disconnected(_) => {
                    return Ok(DispatchDebugResult::Disconnected)
                }
            };

            match r {
                Ok(v) => self.gdb = Some(v),
                Err(e) => return Err(e),
            }
        }
    }
```

```rust
impl<'a, 'b, H: Hypervisor> CpuManager<'a, 'b, H> {
    pub(super) fn dispatch_gdb_idle(
        &mut self,
        mut state: GdbStubStateMachineInner<
            'static,
            Idle<CpuManager<'a, 'b, H>>,
            CpuManager<'a, 'b, H>,
            DebugClient,
        >,
    ) -> Result<
        Result<
            GdbStubStateMachine<'static, CpuManager<'a, 'b, H>, DebugClient>,
            GdbStubStateMachineInner<
                'static,
                Idle<CpuManager<'a, 'b, H>>,
                CpuManager<'a, 'b, H>,
                DebugClient,
            >,
        >,
        DispatchGdbIdleError,
    > {
        let b = match state.borrow_conn().read() {
            Ok(v) => v,
            Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => return Ok(Err(state)),
            Err(e) => return Err(DispatchGdbIdleError::ReadData(e)),
        };

        state
            .incoming_data(self, b)
            .map(Ok)
            .map_err(DispatchGdbIdleError::ProcessData)
    }

    pub(super) fn dispatch_gdb_running(
        &mut self,
        mut state: GdbStubStateMachineInner<'static, Running, CpuManager<'a, 'b, H>, DebugClient>,
        stop: Option<MultiThreadStopReason<u64>>,
    ) -> Result<
        Result<
            GdbStubStateMachine<'static, CpuManager<'a, 'b, H>, DebugClient>,
            GdbStubStateMachineInner<'static, Running, CpuManager<'a, 'b, H>, DebugClient>,
        >,
        DispatchGdbRunningError,
    > {
        // Check If we are here because of a breakpoint.
        if let Some(r) = stop {
            return state
                .report_stop(self, r)
                .map(Ok)
                .map_err(DispatchGdbRunningError::ReportStopReason);
        }

        // Check for pending command.
        let b = match state.borrow_conn().read() {
            Ok(v) => v,
            Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => return Ok(Err(state)),
            Err(e) => return Err(DispatchGdbRunningError::ReadData(e)),
        };

        state
            .incoming_data(self, b)
            .map(Ok)
            .map_err(DispatchGdbRunningError::ProcessData)
    }
}
```

It took several attempt for me to have this working version because we can't do blocking read on the client socket. It should be a lot easier if `gdbstub` change to push model instead of pull model.

ultimaweapon:
> The whole intent behind the `incoming_data` API, the state machine as a whole, and the distinction between [`Connection`](https://docs.rs/gdbstub/latest/gdbstub/conn/trait.Connection.html) and [`ConnectionExt`](https://docs.rs/gdbstub/latest/gdbstub/conn/trait.ConnectionExt.html) was to allow async reads from the GDB client socket.

I think push-model don't forbid this. You have an async event loop that `await` for some data from the client socket and push the received data to the `gdbstub`.

> To give an example: in bare-metal `gdbstub` integrations, folks will typically create a global `GdbStubStateMachine` object, which is then accessed and driven by two independent event sources: e.g: a UART IRQ handling incoming data from the GDB client, and then the software/hardware breakpoint IRQ handler in order to report stop reasons.

I also think push-model does not forbid this. People still able to have a global object and push the received data or interrupt from both location.

> I'm not familiar with the specifics of your integration / async model, but it _should_ be entirely possible to accomplish what you're asking for with the current API. And if not - please let me know where the gap is.

Yes it is possible. It just hard to use and not ergonomic in our case since our event loop waiting for an external event like readable on GDB socket or the VM hit a breakpoint. Once we receive those event we still need to drive the `GdbStubStateMachine` to the point where it ask for data or stop reason. This state machine is easy to use only if we spawn a dedicated thread to drive it and use a blocking I/O but it feel awkward when integrating it into an async event loop.

ultimaweapon:
> Spawning a thread which uses async / epoll to match on incoming socket read / VM stop requests is certainly easier, but theoretically, you can put the GdbStub behind a Mutex and avoid that extra thread to coordinate?

I don't really like to spawn a dedicated thread since integrating its into the same event loop is much more elegance.

> If you could sketch out what an "ideal" API might look like, that might make it easier to get on the same page here. I suspect skimming through pseudocode will be better than trying to explain using just words what API you're looking for.

Good idea. Here is the API that easier to use for our case:

```rust
impl GdbStub {
    pub fn data_received(&mut self, target: &mut impl Target, data: &[u8]);
    pub fn handle_stop(&mut self, target: &mut impl Target, reason: MultiThreadStopReason<u64>);
}
```

With this API we don't need to drive the state machine to the point where it ask for more data. We just push the data we received from the client and stop reason when our VM hit a breakpoint.

> Unless you're concerned about having to use blocking IO within various Target handler methods? e.g: needing to block in order to read/write registers/ram? This is indeed a known limitation, which is currently being tracked in [daniel5151/gdbstub#159](https://github.com/daniel5151/gdbstub/issues/159)

We don't have this kind of problem and I think we will never have.

daniel5151:
Some thoughts:

The simple-looking API you are proposing is incompatible with `gdbstub`'s core mantra of using the Rust type-system to enforce protocol invariants at compile time. The reason there is a state machine with dedicated type-states in the first place is to enforce certain invariants, such as "don't call `handle_stop` if the stub is in the stopped/disconnected state" and "a Disconnected stub is ineligible to receive more data".

> With this API we don't need to drive the state machine to the point where it ask for more data

This is a misunderstanding of how the GDB RSP works, at least, in the world where you're not using https://sourceware.org/gdb/current/onlinedocs/gdb.html/Non_002dStop-Mode.html (which isn't something `gdbstub` currently supports, and moreover, doesn't seem to be supported by most home-grown GDB stub implementations).

In regular GDB operation, the moment you've reported a stop reason to `gdbstub`, every single vCPU in your VM needs to be entirely paused. While paused, the only events you should expect to receive are from the GDB client, which sends you packets that determine what sorts of operations you should do. Eventually, GDB will send you a "resume" packet, at which point, `gdbstub` will ask you to restart your vCPUs, and transition into the "Running" state, such that the next time you get a event (be it new data from the GDB socket, or a stop reason from your VM), you can go back to step 0 and hand control over the gdbstub.

All that is to say, regardless if you are using a type-safe or non-type-safe API to drive the state machine... _someone_ will still need to drive the state machine - be it a dedicated thread / async task, or you manually stepping the machine from various disjoint handler methods across multiple threads.

With the API you proposed, the only way for it to work would be if `GdbStub` became some kind of "thin handle" to a channel / mutex that then coordinated requests to some kind of GdbStub "driver" task / thread, which blocked-on receiving data / stop-reasons via your proposed APIs.

Does that help clarify things? Or am I still missing something obvious 😅

ultimaweapon:
Thanks for explain how it works. I already understand in the same way as you described.

daniel5151:
Cool. If that's the case... do you have any asks here?

Is the key qualm you have with `gdbstub` one of capabilities impossible with the current API, or annoyance with how un-ergonomic it is to drive the underlying state machine directly?

If its the latter, hopefully some of my explanation above clarified why the API is as "complicated" as it currently is, and the rationale behind not having an in-box "helper" to make it easier to interface with it in an async matter (i.e: `gdbstub` hasn't been in the business of favoring any particular async paradigm - be it epoll, async, etc... - but that's not to say I'm opposed to adding some in-box helpers behind a feature-flag)

ultimaweapon:
> or annoyance with how un-ergonomic it is to drive the underlying state machine directly?

This one.

I understand your reason why the API was designed like that. The problem is our problem still not solved. When I choose to use an external library for a specific task it is because I don't want to deal with the underlying technology. All I want is easy to use API. For example, I use Tokio because I want an easy to use high-performance async I/O API. I don't care how it is using epoll/kqueue/IOCP under the hood. Its job is to hide those complexity from the user and expose easy to use API.

daniel5151:
Fair enough! I really appreciate all the feedback and back-and-forth here.

`gdbstub`'s primary focus is to hide the complexity of the underlying GDB RSP protocol from the end user, and in that sense, I think its fair to say it does an outstanding job. If you attempt to write your own stub that processes individual packets... get ready for a world of fun edge cases and complexity 😅

That being said, `gdbstub` is still beholden to the broad constraints of the protocol itself, and as such, pushes some amount of complexity to the end user. This is obvious stuff like "tell me how to access your target", but also non-obvious stuff like "tell me how you wish to interface GDB with your implementation". Helpers can go a long way to reducing the pain... but there will always be some complexity end-users will have to grapple with.

I've tried to offer a few helpers in-box myself, but given the wide variety of paradigms folks use to model their event loops, I haven't spent cycles _personally_ auditing those various paradigms / integrating with those various libraries directly. Maybe a few years ago, when I had more free time, that could've been fun to do... but life is a bit busy at the moment, and I've relied on contributors to `gdbstub` to take point on these sorts of "useful for my paradigm" class of features.

If you end up finding a nice abstraction around the state machine API that you think would benefit other folks trying to integrate `gdbstub` into a similar async system, feel free to submit a PR my way.

ultimaweapon:
Thanks for reaching out and trying to understand our problems even my English is quite bad. I appreciate your works which I believe it is useful to a lot of people out there.

daniel5151:
Thank you! And no need to sell yourself short - your English is reasonable, at least in the context of this particular technical discussion :)

PR #1295, opened 2025-02-28
Upgrades to Slint 1.10

ultimaweapon:
This version come with menu bar. Combined this with our modal window we can get rid of our own custom menu page.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,6 +54,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "aligned-vec"
+version = "0.5.0"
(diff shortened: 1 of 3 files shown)


SuchAFuriousDeath:
The file it's looking for really doesn't exist. There is one with a similar name, but it has an extra -svg- in its name.

ultimaweapon:
I think for Skia on Windows better to use DirectX instead of Vulkan so it can use a pre-built binary instead of building from source.

SuchAFuriousDeath (approved):

PR #1323, opened 2025-04-05
Implements set_center for X11

SuchAFuriousDeath:


SuchAFuriousDeath:
I'm not sure if assuming setup.roots().next() will always return some is ok.

ultimaweapon (approved):
I can review only the code but can't verify if it is working since I don't use X11.

PR #1333, opened 2025-04-27
Skips pmap_bootstrap and spawn_pagers implementation for now

ultimaweapon:
Too many unknown variables. We will come back to implement it later once we reach the code that using those variables.


lib/krt/src/console/mod.rs
```diff
@@ -86,9 +86,32 @@ impl<M: Display> Display for Log<'_, M> {
...
         Ok(())
     }
 }
+
+/// Struct to indent multi-line message.
+struct MsgWriter<'a, 'b>(&'a mut Formatter<'b>);
+
+impl Write for MsgWriter<'_, '_> {
+    fn write_str(&mut self, mut s: &str) -> core::fmt::Result {
+        while let Some(i) = s.bytes().position(|b| b == b'\n') {
```
SuchAFuriousDeath, line 106: Would using https://doc.rust-lang.org/std/primitive.str.html#method.lines be possible here?

SuchAFuriousDeath (changes requested):

lib/krt/src/console/mod.rs
ultimaweapon: Good idea.

SuchAFuriousDeath (approved):

PR #1339, opened 2025-05-05
Finishes vm_page_alloc implementation when available pages lower than v_interrupt_free_min

ultimaweapon:


kernel/src/vm/mod.rs
```diff
@@ -89,13 +95,8 @@ impl Vm {
...
     /// |PS4 11.00|0x02B030|
-    pub fn alloc_page(&self, obj: Option<VmObject>, flags: VmAlloc) {
-        // Get target VM.
-        let vm = match obj {
-            Some(_) => todo!(),
-            None => 0,
-        };
-
+    pub fn alloc_page(&self, obj: Option<VmObject>, flags: VmAlloc) -> Option<VmPage> {
+        let vm = obj.as_ref().map(|v| v.vm()).unwrap_or(0);
```
SuchAFuriousDeath, line 99: This could be simplified with .map_or() I think, but up to you.

lib/bitflag/src/lib.rs
```diff
@@ -15,3 +15,14 @@ pub trait Type: From<Self::Raw> {
 pub trait Raw: Eq + BitOr<Output = Self> + Not<Output = Self> + Copy {}
 
 impl Raw for u32 {}
+
+/// Provides method to construct a value from [`Raw`].
+pub trait FromRaw<T>: Sized {
```
SuchAFuriousDeath, line 20: This just seems like it's basically TryFrom from the standard library?. If you just want a custom trait bound to use for the get method, maybe you could do 
pub trait FromRaw<T>: TryFrom<T> {}

impl FromRaw<u32> for u16 {}
```rs


And then in the get method, just use try_from
```

SuchAFuriousDeath: (Unless I'm missing something)

SuchAFuriousDeath (changes requested):

lib/bitflag/src/lib.rs
SuchAFuriousDeath: Oh, ok then.

SuchAFuriousDeath (approved):

PR #1382, opened 2025-09-06
Applies edited environment to the table

ultimaweapon:


ultimaweapon:
The error on macOS likely to be the same as https://github.com/rust-lang/rust/issues/146239. So let's wait for it to be fixed.

SuchAFuriousDeath (approved):

PR #1397, opened 2025-10-25
Ignores qVAttachOrWaitSupported from GDB

ultimaweapon:
Can't find information how to implement this.


ultimaweapon:
> So I suppose our current response is fine and indicates 'Not supported'.

Probably. Not sure if `vAttachOrWait` useful to us. If yes we probably want to create an issue for this to implement later.

PR #1407, opened 2025-11-16
gdb: Improve byte matching

SuchAFuriousDeath:
This is just to improve readability, no functional change intended. I also reordered the packets alphabetically.


ultimaweapon (approved):
I'm not a fan of using macro like this due to it hide the actual code (e.g. you need to read the macro to understand how the code is working) but let's do this to see if we have benefits from it more than problems.

PR #1409, opened 2025-11-17
gdb: Implement ?, qC, qfThreadInfo, qRegisterInfo and qsThreadInfo

SuchAFuriousDeath:


gui/src/gdb/client.rs
```diff
@@ -101,12 +101,22 @@ impl<'a, H: GdbHandler> GdbDispatcher for ClientDispatcher<'a, H> {
         let off = res.len();
 
         match_bytes! { data,
+            // https://sourceware.org/gdb/onlinedocs/gdb/Stop-Reply-Packets.html
+            "?" => state.parse_stop_reason(res),
```
ultimaweapon: Better to put a link to `?` document itself too since it is hard to find (I'm one of people who can't find it).

gui/src/gdb/state.rs
```diff
@@ -110,4 +112,112 @@ impl SessionState {
...
     }
+
+    pub fn parse_current_thread(&mut self, res: &mut Vec<u8>) {
+        // Return thread ID 1 as the current thread.
+        res.extend_from_slice(b"QC1");
+    }
+
+    pub fn parse_stop_reason(&mut self, res: &mut Vec<u8>) {
+        // Report stopped due to SIGTRAP (signal 5).
+        res.extend_from_slice(b"S05");
```
ultimaweapon: Better to put a link where this number come from.

gui/src/gdb/state.rs
```diff
@@ -110,4 +112,112 @@ impl SessionState {
         // Only Continue and Stop is supported at the moment.
         res.extend_from_slice(b"vCont;c;t");
     }
+
+    pub fn parse_current_thread(&mut self, res: &mut Vec<u8>) {
+        // Return thread ID 1 as the current thread.
+        res.extend_from_slice(b"QC1");
```
ultimaweapon: Better to put a TODO here to retrieve the value from VMM.

ultimaweapon (changes requested):
Good works. Need some comments to make things easier to understand for other people.

ultimaweapon (approved):
Better to not force-push after the review since it make sub-sequence review a lot harder.

